PR #82, opened 2017-11-19
`makeKeyAndVisible` is called twice. It removes the 2nd one.

coybit:


michaelScoff:
@coybit! Thank you for contribution!

michaelScoff:
@coybit do you have telegram? Could you join our telegram group? https://t.me/trustwallet Would be happy to chat

PR #94, opened 2017-11-26
Convert to Swift BigInt library

### LINKED ISSUE #90: Convert into using BigInt library
https://github.com/attaswift/BigInt

aleph7:
Fix #90


Trust/Models/ParsedTransaction.swift
```diff
@@ -76,19 +77,19 @@ extension ParsedTransaction {
         let nonce = transaction["nonce"] as? String ?? "0"
         return ParsedTransaction(
             blockHash: blockHash,
-            blockNumber: BInt(hex: blockNumber.drop0x).dec,
-            transactionIndex: BInt(hex: transactionIndex.drop0x).dec,
+            blockNumber: BigInt(blockNumber.drop0x, radix: 16)!.description,
```
michaelScoff: Is there a way to remove force unwrap? I feel like if wrong format being returned it could easily crash the app

Trust/Transactions/Types/TransactionCost.swift
```diff
@@ -1,27 +1,34 @@
...
 
-    var gasPrice: GethBigInt {
+    var gasPrice: BigInt {
         switch self {
-        case .fast: return GethNewBigInt(6000000000)
-        case .regular: return GethNewBigInt(2400000000)
-        case .cheap: return GethNewBigInt(1000000000)
-        case .custom(let gasPrice, _): return gasPrice
+        case .fast:
+            return 6_000_000_000
```
michaelScoff: Is this Swift specific format? I haven't seen this before. looks pretty awesome though

Trust/Transactions/ViewModels/TransactionViewModel.swift
```diff
@@ -74,8 +75,8 @@ struct TransactionViewModel {
     }
 
     var gasFee: String {
-        let gasUsed = BInt(String(transaction.gasUsed))
-        let gasPrice = BInt(String(transaction.gasPrice))
+        let gasUsed = BigInt(String(transaction.gasUsed))!
```
michaelScoff: Same question about force unwrap, it would be good to avoid as much as possible.

michaelScoff:
screen shot 2017-11-26 at 12 27 14 am
😍

Trust/Transactions/Types/TransactionCost.swift
aleph7: Yes, makes it easier to read.

PR #104, opened 2017-11-30
Implement new ether number formatter

### LINKED ISSUE #98: Consolidate amount value
Related to https://github.com/TrustWallet/trust-wallet-ios/issues/96

It would be good to have consistent amount string in the app, I propose to have two types:

`1.123` - short type, as regular amount
`1.12312312321222` - long type, full amount

aleph7:
Fix #98


TrustTests/Foundation/EtherNumberFormatterTests.swift
```diff
@@ -41,7 +41,7 @@ class EtherNumberFormatterTests: XCTestCase {
     func testDigits() {
         let number = BigInt("1234567890123456789012345678901")!
         XCTAssertEqual(fullFormatter.string(from: number), "1,234,567,890,123.456789012345678901")
-        XCTAssertEqual(shortFormatter.string(from: number), "1,234,567,890,123.456")
+        XCTAssertEqual(shortFormatter.string(from: number), "1,234,567,890,123.457")
```
michaelScoff: @aleph7  So if a user has 0.01 on his account, but it is actually 0.0099, in that case, he won't be able to send the payment. Would it be possible to avoid rounding? or what's the best solution here?

aleph7: Only the `short` formatter would round, the `full` one will never round.

michaelScoff:
@aleph7 could this be reusable as well with tokens? since they take `decimals` parameter to calculate `Double` value.

PR #114, opened 2017-12-07
Correct sshd path in jailbreak detection code

superduper:
Hi!

This patch fixes `sshd` path at [DeviceChecker.swift:L15](https://github.com/superduper/trust-wallet-ios/blob/master/Trust/Core/Helpers/DeviceChecker.swift#L15).

Best,
Viktor


michaelScoff:
Nice catch @superduper! Thanks

PR #117, opened 2017-12-08
Clarifying Import Option Text

sdwolfe32:
I was trying to import an ETH wallet today and noticed it said "Google Cloud" instead of "Google Drive". Then I realized it only imports from local, so figured I'd help clarify that option.

### DIFF
--- a/Trust/Wallet/ViewControllers/ImportWalletViewController.swift
+++ b/Trust/Wallet/ViewControllers/ImportWalletViewController.swift
@@ -197,7 +197,7 @@ class ImportWalletViewController: FormViewController {
     @objc func importOptions(sender: UIBarButtonItem) {
         let alertController = UIAlertController(title: "Import Wallet Options", message: .none, preferredStyle: .actionSheet)
         alertController.popoverPresentationController?.barButtonItem = sender
-        alertController.addAction(UIAlertAction(title: "Local", style: .default) { _ in
+        alertController.addAction(UIAlertAction(title: "iCloud/Dropbox/Google Drive", style: .default) { _ in
             self.showDocumentPicker()
         })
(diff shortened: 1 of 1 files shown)


michaelScoff:
@sdwolfe32 `Local` sounds vague. Can we think of a better name though? it should be very intuitive
`From Documents`? 

What others apps do?

michaelScoff:
@sdwolfe32 `UIDocumentPickerViewController` support import from apps that you have installed, like Google Drive, iCloud, Dropbox and etc.

michaelScoff:
This is the options I see since I only have iCloud and Google Drive installed

image 2017-12-07 22 32 49

michaelScoff:
@sdwolfe32 also noticed in that file that cancel button had destructive style, so I changed it: https://github.com/TrustWallet/trust-wallet-ios/commit/7756c4a0700f05f31f9c4591ff6d63e1dde99872

michaelScoff:
@sdwolfe32 I would propose to change name fix name as you mentioned to `Google Drive` and add extra etc? In that case it would be more intuitive for that user

sdwolfe32:
@michaelScoff Ahh I see. Sorry, didn't realize you could also access Google Drive and other apps. I'll update to change the wording to "Google Drive" as you said. Sorry for my naiveness.

Thanks! Love the work you're doing!

michaelScoff:
@sdwolfe32 thank you for the first contribution? 

Would you like to help? Join me on slack: https://join.slack.com/t/trust-wallet/shared_invite/enQtMjc1NDIzNTc1NzgxLTg1M2U5N2MzNmVlOTBhYTM4NDMwZDM2OTg1ZDA4NmU2OWFiNTZlN2Q4YmNmZGViYTE5ZTAyM2I3YTZlNjYxYjE

PR #128, opened 2017-12-10
Update App Store description

michaelScoff:
Since we plan to support localization in the app, I would like to have some description for our users, this will be localized as well. Let's make sure it's in a good shape.

Feel free to add and comment on this description

this description will be used for Android as well


michaelScoff:
What can we do to impress users?

PR #137, opened 2017-12-16
Remove short formatter rounding

michaelScoff:


Trust/Foundation/EtherNumberFormatter.swift
```diff
@@ -130,7 +130,6 @@ final class EtherNumberFormatter {
         }
         if decimals > maximumFractionDigits {
             number /= BigInt(10).power(decimals - maximumFractionDigits - 1)
```
aleph7: You also need to remove the `- 1`

michaelScoff: I would like to have 4 digits after `.` is it should be still there?

aleph7: ok

PR #148, opened 2017-12-20
Added params parsing support to QR parser

KamuelaFranco:


Trust/Foundation/QRURLParser.swift
```diff
@@ -8,6 +8,23 @@ struct ParserResult {
     let params: [String: String]
 }
 
+func parseParamsFromParamParts(paramParts: [String]) -> [String: String] {
+    if paramParts.isEmpty {
```
michaelScoff: we could use guard here, usually easier: `guard !paramParts.isEmpty else { return [:] }`

Trust/Foundation/QRURLParser.swift
```diff
@@ -8,6 +8,23 @@ struct ParserResult {
     let params: [String: String]
 }
 
+func parseParamsFromParamParts(paramParts: [String]) -> [String: String] {
```
michaelScoff: Could you write tests to make sure it's working properly? check out `QRURLParserTests`

PR #151, opened 2017-12-20
Memory leaks in SendViewController.

### LINKED ISSUE #150: Memory leaks in SendViewController.
During long use of the application SendViewController is not deallocated.

OlegGordiichuk:
I remove memory leaks that are in SendViewController via usage of [weak self] instead of self in closures.

Fix #150


michaelScoff:
Does this the only place that has memory leak? It would be good to go over and check. 
We could probably just use unowned self instead.

OlegGordiichuk:
@michaelScoff unowned is it quite expensive take a look at https://www.uraimo.com/2016/10/27/unowned-or-weak-lifetime-and-performance/

michaelScoff:
Agreed that unowned could be more expensive, but it makes it harder to reason behavior of the current code. Eureka won’t be calling call update or hidden function once view controller is hidden though. I don’t think it would be a big penalty

Do you mind helping to going through the code to find more places that has it? I might have missed in lot of places, I will be more aware now though.

PR #154, opened 2017-12-20
We should clean coordinator stack in restart function.

### LINKED ISSUE #153: Memory leak in network change.
We receive huge memory leaking after changing of the network type.

OlegGordiichuk:
We should always clear navigation stack on the restart to avoid zombie objects.

Also could you pls validate if all steps with navigation are working fine. From my side I check this but double check I think is required.

Fix #153


michaelScoff:
@OlegGordiichuk looks good! we could even add tests to make sure it's removed coordinators on those actions

PR #156, opened 2017-12-20
Welcome screen.

### LINKED ISSUE #155: We have always WelcomeViewController in the navigation stack.
We should show WelcomeViewController when we do not have any account or we log out.

OlegGordiichuk:
We should show welcome screen if we do not have account not before validation of the account.

Fix #155

### DIFF
--- a/Trust/AppCoordinator.swift
+++ b/Trust/AppCoordinator.swift
@@ -43,6 +43,8 @@ class AppCoordinator: NSObject, Coordinator {
 
         if keystore.hasAccounts {
             showTransactions(for: keystore.recentlyUsedAccount ?? keystore.accounts.first!)
+        } else {
+            resetToWelcomeScreen()
         }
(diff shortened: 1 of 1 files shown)


michaelScoff:
We also have a call `resetToWelcomeScreen()` on line 42, would it make sense to call it once?

michaelScoff:
if you log out then it would call `resetToWelcomeScreen()` I believe

OlegGordiichuk:
Hmmm we have two scenarios that are:
1. Login.
    

- We should show Welcome screen only when we do not have **keystore.hasAccounts**.
- Now we show Welcome screen and than do validation. (**Problem is in this step**). That Welcome     screen is always in stack.

2. Log out (We should show Welcome screen without any condition.)

Yes you are right we should show WelcomeScreen preferably only once and do not use if statements. But for now I think it is possible to apply this fix. And create technical debt issue for proper navigation flow for different scenarious.

PR #176, opened 2017-12-29
Replace code block with offset to viewWillAppear.

### LINKED ISSUE #162: Status bar squished on iPad
GitHawk Upload by michaelScoff

OlegGordiichuk:
Fix #162

### DIFF
--- a/Trust/Transactions/ViewControllers/TransactionsViewController.swift
+++ b/Trust/Transactions/ViewControllers/TransactionsViewController.swift
@@ -28,6 +28,8 @@ class TransactionsViewController: UIViewController {
     weak var delegate: TransactionsViewControllerDelegate?
     let dataCoordinator: TransactionDataCoordinator
     let session: WalletSession
+    
+    let insets = UIEdgeInsets(top: 130, left: 0, bottom: ButtonSize.extraLarge.height + 84, right: 0)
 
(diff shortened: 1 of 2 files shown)


Trust/Transactions/ViewControllers/TransactionsViewController.swift
```diff
@@ -77,9 +79,12 @@ class TransactionsViewController: UIViewController {
...
         tableView.addSubview(refreshControl)
 
-        //TODO: Find a way to fix hardcoded 32px value. Use bottom safe inset instead.
-        let insets = UIEdgeInsets(top: 130, left: 0, bottom: ButtonSize.extraLarge.height + 84, right: 0)
+        navigationItem.titleView = titleView
+        titleView.viewModel = BalanceViewModel()
+    }
 
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
```
michaelScoff: I would expect it was a bug because `super.viewWillAppear(animated)` wasn't called?

do we need to set insets every time to solve this issue

OlegGordiichuk: I update method.

OlegGordiichuk:
I fix core issue that was causing problems.

PR #181, opened 2018-01-03
Updated social sharing text

BorisButakov:


michaelScoff:
@BorisButakov can you add additionally to the first two? and I will merge then, 4 in total.

PR #228, opened 2018-01-18
Change my address

James-Sangalli:
This PR changes My Address to receive. This prevents confusion.


michaelScoff:
I'm not convinced it should be called `Receive` yet, if so it needs to be changed on `lokalise`. 

I will close this for now, until we get to this part of UX

PR #235, opened 2018-01-21
Issues/234

### LINKED ISSUE #234: Rewrite of the passlock behavior.

OlegGordiichuk:
Fix #234


Trust/Protection/Coordinators/SplashCoordinator.swift
```diff
@@ -0,0 +1,23 @@
...
+import UIKit
+
+class SplashCoordinator: Coordinator {
+    var coordinators: [Coordinator] = []
+    let navigationController: UINavigationController
+    private lazy var splashViewController: SplashViewController = {
+        return SplashViewController()
+    }()
+    init(
+        navigationController: UINavigationController = NavigationController()
```
michaelScoff: `SplashCoordinator` should be showing `UIWindow` that's separate from the main window, in that case, it won't affect a state of the current application.

Trust/Core/Initializers/TouchRegistrar.swift
```diff
@@ -28,11 +28,8 @@ class TouchRegistrar {
             keychainAccount: Keys.account,
             touchIDReason: "Use your fingerprint to access your wallet",
             passcodeAttemptLimit: 5,
-            splashViewControllerClass: SplashViewController.self
+            splashViewControllerClass: ProtectionViewController.self
```
michaelScoff: Let's not use Protection screen from VENTouch at all.

PR #255, opened 2018-01-28
Issues/238

OlegGordiichuk:
Fix #255


Podfile
```diff
@@ -18,16 +18,16 @@ target 'Trust' do
...
   pod 'Lokalise'
   pod 'Moya', '~> 10.0.1'
   pod 'JavaScriptKit' 
   pod 'CryptoSwift', :git=>'https://github.com/krzyzanowskim/CryptoSwift', :branch=>'master'
   pod 'Fabric'
+  pod 'SAMKeychain'
   pod 'Crashlytics'
   pod 'Kingfisher', '~> 4.0'
-  pod 'TrustKeystore', :git=>'https://github.com/TrustWallet/trust-keystore', :branch=>'master'
+  pod 'TrustKeystore', '~> 0.0.4'
```
michaelScoff: Can we keep this separate? I think we need 0.0.5 version, which is master.

Trust/Lock/Lock.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+    func setPasscode(passcode: String) {
+        SAMKeychain.setPassword(passcode, forService: Keys.service, account: Keys.account)
+    }
+    func deletePasscode() {
+        SAMKeychain.deletePassword(forService: Keys.service, account: Keys.account)
+        resetPasscodeAttemptHistory()
+    }
+    func numberOfAttempts() -> Int {
+        return standardDefaults.integer(forKey: passcodeAttempts)
```
michaelScoff: Can we move this into keychain?

Trust/Lock/Lock.swift
```diff
@@ -0,0 +1,53 @@
...
+    func recordIncorrectPasscodeAttempt() {
+        var numberOfAttemptsSoFar = standardDefaults.integer(forKey: passcodeAttempts)
+        numberOfAttemptsSoFar += 1
+        standardDefaults.set(numberOfAttemptsSoFar, forKey: passcodeAttempts)
+    }
+    func recordedMaxAttemptTime() -> Date {
+        return standardDefaults.object(forKey: maxAttemptTime) as! Date
+    }
+    func incorrectMaxAttemptTimeIsSet() -> Bool {
+        return (standardDefaults.object(forKey: maxAttemptTime) != nil)
```
michaelScoff: Same here, moving into keychain

Trust/Protection/Coordinators/SplashCoordinator.swift
```diff
@@ -4,20 +4,24 @@ import UIKit
...
+        guard !splashViewIsActive else {
+            return
+        }
+        splashViewIsActive = true
+        window.rootViewController = splashViewController
+        window.makeKeyAndVisible()
     }
     func dismiss() {
-        navigationController.popViewController(animated: false)
+        splashViewIsActive = false
```
michaelScoff: `splashViewIsActive` - I think we can check if this is true by: `splashViewIsActive = !window.isHidden`

Trust/Lock/ViewControllers/LockPasscodeViewController.swift
```diff
@@ -0,0 +1,135 @@
+// Copyright SIX DAY LLC. All rights reserved.
+
+import UIKit
+
+class LockPasscodeViewController: UIViewController {
+    var willFinishWithResult: ((_ success: Bool) -> Void)?
+    let model: LockViewModel
+    var lockView: LockView!
```
michaelScoff: can you configure `lockView` as lazy initializer? to remove force unwrap

Trust/Lock/Coordinators/LockEnterPasscodeCoordinator.swift
```diff
@@ -0,0 +1,34 @@
...
+    func start() {
+        guard !passcodeViewIsActive && Lock().isPasscodeSet() else {
+            return
+        }
+        passcodeViewIsActive = true
+        window.rootViewController = lockEnterPasscodeViewController
+        window.makeKeyAndVisible()
+    }
+    func dismiss() {
+        passcodeViewIsActive = false
```
michaelScoff: Same here 
```
var passcodeViewIsActive: bool { return !window.isHidden }
```

Trust/AppDelegate.swift
```diff
@@ -7,6 +7,7 @@ import Lokalise
 class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDelegate {
     var window: UIWindow?
     var coordinator: AppCoordinator!
+    var protectionCoordinator: ProtectionCoordinator!
```
michaelScoff: can we do lazy initializer to protectionCoordinator?

Podfile
OlegGordiichuk: Done I use previous declaration of the app version.

Trust/AppDelegate.swift
OlegGordiichuk: Done

Trust/Lock/ViewControllers/LockPasscodeViewController.swift
OlegGordiichuk: In case of the migration lock view will be installed with 4 digit and when user will decide to change it to 6 digit lazy var will not be updated.

Trust/Lock/Coordinators/LockEnterPasscodeCoordinator.swift
OlegGordiichuk: We have same protection window for splash and lock screen we should strictly define witch screen is presented. But yes I think we could optimize this later.

Trust/Protection/Coordinators/SplashCoordinator.swift
OlegGordiichuk: We have same protection window for splash and lock screen we should strictly define witch screen is presented. But yes I think we could optimize this later.

Trust/Lock/Lock.swift
OlegGordiichuk: Done

Trust/Lock/Lock.swift
OlegGordiichuk: Done

PR #264, opened 2018-02-01
Show gas fee in current currency

hewigovens:
Hi

I add a helper extension to CurrencyRate to keep the same format in two views

Thanks

### DIFF
--- a/Trust/Transfer/ViewControllers/ConfigureTransactionViewController.swift
+++ b/Trust/Transfer/ViewControllers/ConfigureTransactionViewController.swift
@@ -13,6 +13,7 @@ class ConfigureTransactionViewController: FormViewController {
     let configuration: TransactionConfiguration
     let config: Config
     let transferType: TransferType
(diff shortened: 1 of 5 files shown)


Trust/Transfer/ViewControllers/ConfigureTransactionViewController.swift
```diff
@@ -140,7 +143,14 @@ class ConfigureTransactionViewController: FormViewController {
     }
 
     func recalculateTotalFee() {
-        totalFeeRow?.value = "\(fullFormatter.string(from: totalFee)) \(config.server.symbol)"
+
+        let fee = fullFormatter.string(from: totalFee)
+        var feeAndSymbol = "\(fee) \(config.server.symbol)"
+        if let feeInCurrency = currencyRate?.estimate(fee: fee, with: config.server.symbol) {
```
michaelScoff: Would it possible to move this into view model?

PR #266, opened 2018-02-01
Add window level and manually call viewWillAppear(true) to be sure th…

OlegGordiichuk:
Add window level so the all alerts will be under protection screen.
Manually call viewWillAppear to be sure that keyboard and touch validation will be called always if passcode view was shown previously.


Trust/Lock/Coordinators/LockEnterPasscodeCoordinator.swift
```diff
@@ -22,7 +24,9 @@ class LockEnterPasscodeCoordinator: Coordinator {
         }
     }
     func start() {
-        guard !passcodeViewIsActive && lock.isPasscodeSet() else {
+        guard lock.isPasscodeSet() else {
+             //Due to the lazy init and navigation flow we should call viewWillAppear manually.
+            lockEnterPasscodeViewController.viewWillAppear(true)
```
michaelScoff: I would personally call `start()` or `activate()` for clarity.

Is this long-term fix? we can wait probably

PR #267, opened 2018-02-02
remove -- and fix alignment for edit tokens

MillerApps:
Removed the -- on the token page as well as fixed alignment of ETN label on the edit tokens section.

### DIFF
--- a/Trust/Tokens/Views/EditTokenTableViewCell.swift
+++ b/Trust/Tokens/Views/EditTokenTableViewCell.swift
@@ -22,7 +22,12 @@ class EditTokenTableViewCell: UITableViewCell {
             tokenLabel.font = viewModel.titleFont
             tokenLabel.textColor = viewModel.titleTextColor
             tokenEnableSwitch.isOn = viewModel.isEnabled
-            tokenContractLabel.text = viewModel.contractText
+            if viewModel.contractText == nil {
+                tokenContractLabel.isHidden = true
+                tokenContractLabel.removeFromSuperview()
+            } else {
+                tokenContractLabel.text = viewModel.contractText
+            }
             tokenEnableSwitch.isHidden = viewModel.isSwitchHidden
             tokenImageView.kf.setImage(
                 with: viewModel.imageUrl,
--- a/Trust/Tokens/Views/EditTokenTableViewCell.swift
+++ b/Trust/Tokens/Views/EditTokenTableViewCell.swift
@@ -22,12 +22,8 @@ class EditTokenTableViewCell: UITableViewCell {
             tokenLabel.font = viewModel.titleFont
             tokenLabel.textColor = viewModel.titleTextColor
             tokenEnableSwitch.isOn = viewModel.isEnabled
-            if viewModel.contractText == nil {
-                tokenContractLabel.isHidden = true
-                tokenContractLabel.removeFromSuperview()
-            } else {
-                tokenContractLabel.text = viewModel.contractText
-            }
+            tokenContractLabel.text = viewModel.contractText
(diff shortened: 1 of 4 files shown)


Trust/Tokens/Views/EditTokenTableViewCell.swift
```diff
@@ -22,7 +22,12 @@ class EditTokenTableViewCell: UITableViewCell {
             tokenLabel.font = viewModel.titleFont
             tokenLabel.textColor = viewModel.titleTextColor
             tokenEnableSwitch.isOn = viewModel.isEnabled
-            tokenContractLabel.text = viewModel.contractText
+            if viewModel.contractText == nil {
+                tokenContractLabel.isHidden = true
+                tokenContractLabel.removeFromSuperview()
```
michaelScoff: I don't think it's a good idea to remove view. Can you use Stack View instead? it supposed to center the view if label is empty

michaelScoff: I believe stack view will remove view if it's empty automatically

michaelScoff:
@MillerApps I think initially we need to center labels on tokens view right?

michaelScoff:
@MillerApps do you mind providing screenshot to see what changed?

MillerApps:
@michaelScoff I guess my eyes are a bit off as they looked like they were centered.

MillerApps:
simulator screen shot - iphone 8 plus - 2018-02-02 at 15 03 47
simulator screen shot - iphone 8 plus - 2018-02-02 at 15 03 41

michaelScoff:
I believe Ethereum (ETH) is not centered, you can check by enabling Debug View Hierarchy 

is there a way not to use `removeFromSuperview` call ?

MillerApps:
@michaelScoff Yes, it works without removeFromSuperView.

Trust/Tokens/Views/EditTokenTableViewCell.swift
```diff
@@ -22,7 +22,12 @@ class EditTokenTableViewCell: UITableViewCell {
             tokenLabel.font = viewModel.titleFont
             tokenLabel.textColor = viewModel.titleTextColor
             tokenEnableSwitch.isOn = viewModel.isEnabled
-            tokenContractLabel.text = viewModel.contractText
+            if viewModel.contractText == nil {
+                tokenContractLabel.isHidden = true
```
michaelScoff: can you move `isTokenContractLabelHidden` into view model? Cell does not need to do any logic in this case

michaelScoff:
@MillerApps just double check alignment in debug view and this should be good to go! + previous comment

Trust/Tokens/Views/EditTokenTableViewCell.swift
MillerApps: I can when I get home. I definitely feel a bit rusty, haven't done much iOS work in a good while.

michaelScoff:
@MillerApps since you are doing this change can you set line break for the contract in the middle?

MillerApps:
@michaelScoff I checked the alignment it's a hair off, I'll have to tweak it.

MillerApps:
Yes when I get back home.

michaelScoff:
@MillerApps nice catch! Making app pixel perfect is hard, but we gonna make it!

MillerApps:
Should be all set now.

MillerApps:
FYI Travis gets hung up on installing TrezorCrypto.

michaelScoff:
@MillerApps Nice work! Merging this now!

@aleph7 can you look into Trezor issue?

PR #270, opened 2018-02-03
Add string format to display double value.

OlegGordiichuk:
Add  string format to display double value in send view controller.
simulator screen shot - iphone 8 - 2018-02-03 at 22 31 52


michaelScoff:
Nice work!

michaelScoff:
@OlegGordiichuk can you double check if you can input values from different locale? like French and etc, so it's parsed correctly?

PR #271, opened 2018-02-03
Bug/402

OlegGordiichuk:
Rewrite logic for showing of the passcode. I take a look at apps that are in GitHub and use their approach. Add unit tests. Test on iPhone (10.3) and iPad (11).


Trust/Lock/Coordinators/LockEnterPasscodeCoordinator.swift
```diff
@@ -13,28 +13,27 @@ class LockEnterPasscodeCoordinator: Coordinator {
...
                 self?.stop()
             }
+            self?.protectionWasShown = bioUnlock
         }
     }
     func start() {
         guard lock.isPasscodeSet() else {
-             //Due to the lazy init and navigation flow we should call viewWillAppear manually.
-            lockEnterPasscodeViewController.viewWillAppear(true)
             return
```
michaelScoff: should we be calling stop method if passcode not set?

Trust/Protection/Coordinators/ProtectionCoordinator.swift
```diff
@@ -12,24 +12,25 @@ class ProtectionCoordinator: Coordinator {
     }()
     let protectionWindow: UIWindow
     init() {
+        let protection = UIWindow()
```
michaelScoff: I think you can move initializer on top `let protectionWindow = UIWindow()`

Trust/Protection/Coordinators/SplashCoordinator.swift
```diff
@@ -4,24 +4,25 @@ import UIKit
...
+            splashView.leadingAnchor.constraint(equalTo: window.leadingAnchor),
+            splashView.bottomAnchor.constraint(equalTo: window.bottomAnchor),
+            splashView.trailingAnchor.constraint(equalTo: window.trailingAnchor),
+        ])
     }
-    func dismiss() {
-        splashViewIsActive = false
-        window.isHidden = true
+    func stop() {
+        splashView.removeFromSuperview()
```
michaelScoff: Do you need to hide window as well?

michaelScoff:
Can you add tests for `LockEnterPasscodeCoordinator` where user disable or enable during using the app. Just want to make sure we cover all the cases

Trust/Protection/Coordinators/SplashCoordinator.swift
OlegGordiichuk: No splash screen we add like subview.

TrustTests/Coordinators/SplashCoordinatorTests.swift
```diff
@@ -0,0 +1,22 @@
...
+        coordinator.start()
+        XCTAssertFalse(window.isHidden)
+        XCTAssertNotNil(window.subviews.first)
+    }
+    func testStop() {
+        let window = UIWindow()
+        let coordinator = SplashCoordinator(window: window)
+        coordinator.start()
+        coordinator.stop()
+        XCTAssertNil(window.subviews.first)
```
michaelScoff: Can you check if it's hidden as well on stop? Of if it doesn't have subviews than that's same thing?

TrustTests/Coordinators/SplashCoordinatorTests.swift
```diff
@@ -0,0 +1,22 @@
...
+@testable import Trust
+
+
+class SplashCoordinatorTests: XCTestCase {
+    func testStart() {
+        let window = UIWindow()
+        let coordinator = SplashCoordinator(window: window)
+        coordinator.start()
+        XCTAssertFalse(window.isHidden)
+        XCTAssertNotNil(window.subviews.first)
```
michaelScoff: One thing to check levels as well?

PR #276, opened 2018-02-04
Fix name saving when creating new tokens

MillerApps:
Fixed name saving when creating new tokens.


michaelScoff:
@MillerApps nice catch!

PR #278, opened 2018-02-04
Code cleaning and removal of the memory leaking.

OlegGordiichuk:
Code cleaning and remove memory leaks.

### DIFF
--- a/Trust/InCoordinator.swift
+++ b/Trust/InCoordinator.swift
@@ -189,6 +189,7 @@ class InCoordinator: Coordinator {
     }
 
     func restart(for account: Wallet, in coordinator: TransactionCoordinator) {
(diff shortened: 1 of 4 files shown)


Trust/InCoordinator.swift
```diff
@@ -189,6 +189,7 @@ class InCoordinator: Coordinator {
     }
 
     func restart(for account: Wallet, in coordinator: TransactionCoordinator) {
+        self.navigationController.dismiss(animated: false, completion: nil)
         coordinator.navigationController.dismiss(animated: true, completion: nil)
```
michaelScoff: do we need to cancel coordinator if current navigation dismissed?

OlegGordiichuk: We should just clean local navigationController.

PR #287, opened 2018-02-06
Add GasViewModel

hewigovens:
Encapsulate gas fee in currency calculations into the  GasViewModel


Trust/Core/ViewModels/GasViewModel.swift
```diff
@@ -0,0 +1,33 @@
...
+struct GasViewModel {
+    let fee: BigInt
+    let symbol: String
+    let currencyRate: CurrencyRate?
+    let formatter: EtherNumberFormatter
+
+    init(
+        fee: BigInt,
+        symbol: String,
+        currencyRate: CurrencyRate? = nil,
```
michaelScoff: I would like it as required, to make sure developer aware of this variable

hewigovens: Ok, I will remove the optional in future commits

PR #291, opened 2018-02-07
Let Dapps Detect TrustWallet

mishfit:
Just dipping my toes in to this project...a small gripe I have with Cipher Browser right now. Particularly since MetaMask uses a different web3 version than mist which probably uses a different version than Cipher Browser.

- MetaMask adds `web3.currentProvider.isMetaMask`
- Mist adds `global.mist`

It would be nice to let dapps easily detect the browsers
in which they're running.


michaelScoff:
@mishfit Niceee! should we add `global.mist` as well then?

mishfit:
I like `isTrust`. I don't think there's a need to add the other properties. It's just a good way to tell that async/await won't work with Mist (I haven't tested Cipher Browser yet).

PR #292, opened 2018-02-07
Remove splash screen on applicationWillEnterForeground

michaelScoff:
```
applicationWillEnterForeground

This method is called as an app is preparing to move from the background to the foreground. The app, however, is not moved into an active state without the applicationDidBecomeActive method being called. This method gives a developer the opportunity to re-establish the settings of the previous running state before the app becomes active.
```


OlegGordiichuk:
In version 10.2 all if working fine. Still need more precise testing.

PR #297, opened 2018-02-08
Add send value validation and replacing of the , to .

### LINKED ISSUE #296: Handle command and dot for input values.
If you are using UK keyboard you won't be able to use `.`, or if in US then `,`.
We need to have a parser that will try to parse locale if that fails, try different ways.

Let's create separate parser that we could test.

OlegGordiichuk:
Add logic to support validation of the decimal separator.

Fix #296

### DIFF
--- a/Trust/Foundation/DecimalFormatter.swift
+++ b/Trust/Foundation/DecimalFormatter.swift
@@ -5,17 +5,37 @@ import Foundation
 final class DecimalFormatter {
     /// Locale of a `DecimalFormatter`.
     var locale: Locale
-    /// numberFormatter of a `DecimalFormatter`.
-    private var numberFormatter: NumberFormatter
-    /// usLocale of a `DecimalFormatter` to represent decimal separator ".".
-    private let usLocale = Locale(identifier: "en_US")
-    /// usLocale of a `DecimalFormatter` to represent decimal separator ",".
-    private let frLocale = Locale(identifier: "fr_FR")
-    /// enCaLocale of a `DecimalFormatter` to represent decimal separator "'".
-    private let enCaLocale = Locale(identifier: "en_CA")
-    /// locales of a `DecimalFormatter` to check locale inconsistency.
-    private lazy var locales: [Locale] = {
-        return [locale, usLocale, frLocale, enCaLocale]
+    /// numberFormatter of a `DecimalFormatter` to represent curent locale.
+    private lazy var numberFormatter: NumberFormatter = {
+        let formatter = NumberFormatter()
+        formatter.locale = self.locale
+        formatter.numberStyle = .decimal
+        formatter.isLenient = true
+        return formatter
+    }()
+    /// usFormatter of a `DecimalFormatter` to represent decimal separator ".".
+    private lazy var usFormatter: NumberFormatter = {
+        let formatter = NumberFormatter()
+        formatter.locale = Locale(identifier: "en_US")
+        formatter.numberStyle = .decimal
+        formatter.isLenient = true
+        return formatter
+    }()
+    /// frFormatter of a `DecimalFormatter` to represent decimal separator ",".
+    private lazy var frFormatter: NumberFormatter = {
+        let formatter = NumberFormatter()
+        formatter.locale = Locale(identifier: "fr_FR")
+        formatter.numberStyle = .decimal
+        formatter.isLenient = true
+        return formatter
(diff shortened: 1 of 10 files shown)


Trust/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -137,31 +131,27 @@ class SendViewController: FormViewController {
...
-
         let addressString = addressRow?.value?.trimmed ?? ""
         var amountString = ""
         if self.currentPair.left == viewModel.symbol {
             amountString = amountRow?.value?.trimmed ?? ""
         } else {
             amountString = String(format: "%f", self.pairValue).trimmed
         }
-
+        amountString = amountString.validateDecimalSeparator()
```
aleph7: This will not work for strings like `1,001.25`

aleph7: Also we should have unit tests for this.

Trust/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -312,7 +287,7 @@ extension SendViewController: QRCodeReaderDelegate {
 extension SendViewController: UITextFieldDelegate {
     func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
         let text = (textField.text as NSString?)?.replacingCharacters(in: range, with: string)
-        guard let total = text, let amount = Double(total) else {
+        guard let total = text, let amount = Double(total.replacingOccurrences(of: ",", with: ".")) else {
```
aleph7: Same as above.

michaelScoff:
@aleph7 do you have any suggestions for handling both cases? I have been receiving some feedback from users, even tho their keystore has `,`, they would still copy paste value with a `.`. Should we have an error that says that you need to use `.` or `,` or finds a way to handle this case for them(I would prefer this way)

michaelScoff:
@aleph7 also we found interesting use case today. I have Vik coin with `decimals: 0`, so if you try to send `0.1` it would fail to parse. It would nice to make it clear to the user somehow.

aleph7:
I think we should be able to parse every one of these: `1000.25`, `1,000.25`, `1.000,25`, `1'000,25`. Just try each one in order until one succeeds or we run out of options.

OlegGordiichuk:
@aleph7 and @michaelScoff how about to add currency field that will respect 1,001.25 ? And in this field do all required validation ?

michaelScoff:
@OlegGordiichuk currency field might be better for the long term, I need to see UX for it, how convenient it would be for the users. 

But for now, just parsing until it finds a match with different locales.

OlegGordiichuk:
@aleph7  and @michaelScoff I update my solution. 

in order to validate input amount in send view controller and prevent app from alert message from EtherNumberFormatter about wrong input. I add helper class that will tray to find possible value that user would like to input. For this I use 3 possible locale that have different decimal separators.

Trust/Foundation/DecimalFormatter.swift
```diff
@@ -0,0 +1,58 @@
...
+        self.numberFormatter.numberStyle = .decimal
+        self.numberFormatter.isLenient = true
+    }
+    /// Converts a String to a `NSumber`.
+    ///
+    /// - Parameters:
+    ///   - string: string to convert.
+    /// - Returns: `NSumber` represenation.
+    func number(from string: String) -> NSNumber? {
+        return self.numberFormatter.number(from: self.validation(for: string))
```
aleph7: This is good, but we should try all three number formatters. Something like `return usFormatter.number(from: string) ?? frFormatter.number(from: string) ?? enCaFormatter.number(from: string)` instead of changing the string which seems like double the work.

OlegGordiichuk:
@aleph7 I clear code and use swift nil collision operators.

Trust/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -152,10 +152,10 @@ class SendViewController: FormViewController {
             amountString = String(format: "%f", self.pairValue).trimmed
         }
         guard let validatedAmountString = decimalFormatter.number(from: amountString)?.stringValue else {
-            return
+            return displayError(error: Errors.invalidAddress)
```
vikmeup: This should return invalid amount input error

PR #308, opened 2018-02-08
[WIP] Web3 revisited

mishfit:
~~**Do Not Merge Yet**~~

If a dapp uses EthJS, for instance, it's impossible to override interactions with JSON rpc using a simple HttpProvider

***Edit***: good to merge, tested CK


mishfit:
I'd love to get farther than I am on this, but I'm running into:
```
SyntaxError: Invalid escape in identifier: '\'
```
After loading the WKWebView
It must be coming from non-escaped characters in `trust.js` (a browserified web3-provider-engine which should allow us to hook into JSON rpc interactions regardless of a dapp using EthJS or web3)

mishfit:
I've tested `trust.js` and `trust-min.js` separately in Chrome using static-server and that loads just fine

mishfit:
- [x] - signing a transaction
- [x] - sending a transaction _**(via web3, it isn't calling back into swift)**_
- [ ] - signing a message
- [ ] - signing a personal message
- [ ] - signing a typed message

I think once those are done dapps will be satisfied.
Also, how important is it that sending a transaction comes from swift instead of web3?

mishfit:
Ended up not needing `signTypedMessage` just `signPersonalMessage` and `signMessage`

vikmeup:
@mishfit this is amazing! Nice work 👍

Issue #290, opened 2018-02-07
UIBarButtonItem tint should be white on SFSafariViewController view


michaelScoff:
Keep in mind that changing this would cause other issues in the toolbar. Make sure they both handled

![simulator screen shot - iphone se - 2018-02-06 at 20 22 51](https://user-images.githubusercontent.com/32179653/35898426-89f56a28-0b7b-11e8-8e96-4913557ef9f0.png)

PR #309, opened 2018-02-09
fix-tint

MillerApps:
Removed what appears to have been unneeded code which caused the issue mentioned in issue #290

### DIFF
--- a/Trust/Style/AppStyle.swift
+++ b/Trust/Style/AppStyle.swift
@@ -16,9 +16,7 @@ func applyStyle() {
     UINavigationBar.appearance().titleTextAttributes = [
         .foregroundColor: UIColor.white,
     ]
-    UIButton.appearance().tintColor = UIColor.white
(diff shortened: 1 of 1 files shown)


michaelScoff:
Did you check browser view?

MillerApps:
The DApps browser? If so yes.
simulator screen shot - iphone 8 plus - 2018-02-08 at 23 06 16

michaelScoff:
@MillerApps Remember now, do focus on textfield in the send view, it will show input toolbar, those buttons are broken.

MillerApps:
Not sure what you mean they seem to be working for me. 
simulator screen shot - iphone 8 plus - 2018-02-08 at 23 10 08
simulator screen shot - iphone 8 plus - 2018-02-08 at 23 10 02

michaelScoff:
@MillerApps That looks good! Hm, can’t remember then, I will double check tonight if that was it

MillerApps:
It definitely seems like it was but, its always best to check on multiple devices. I'm off to bed now.

PR #310, opened 2018-02-09
added overload to allow signing without a prefix

James-Sangalli:


Trust/EtherClient/EtherKeystore.swift
```diff
@@ -289,8 +289,17 @@ open class EtherKeystore: Keystore {
         }
     }
 
-    func signMessage(message: String, account: Account) -> Result<Data, KeystoreError> {
-        let formattedMessage: String = "\u{19}Ethereum Signed Message:\n" + "\(message.count)" + message
+    func signMessage(_ message: String, _ account: Account, _ prefixAdded: Bool) -> Result<Data, KeystoreError> {
```
vikmeup: Use `prefix: String` as a parameter, can you add tests for it?

Also fix swiftlint errors

vikmeup: or use signPersonal message method.

vikmeup:
Added separate method to `signPersonalMessage`

PR #317, opened 2018-02-13
Fix performance issue in transaction screen.

OlegGordiichuk:
I optimize logic for creating of the sections in the transaction view controller and reload of the table view.

### DIFF
--- a/Trust/Transactions/ViewModels/TransactionsViewModel.swift
+++ b/Trust/Transactions/ViewModels/TransactionsViewModel.swift
@@ -3,6 +3,11 @@
 import Foundation
 import UIKit
 
+struct TransactionsSectionModel {
+    var header: String
+    var items: [Transaction]
+}
+
 struct TransactionsViewModel {
 
     static let formatter: DateFormatter = {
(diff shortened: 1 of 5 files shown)


OlegGordiichuk:
@aleph7 and @michaelScoff  could you pls review logic with performance improvement.

Trust/Transactions/ViewModels/TransactionsViewModel.swift
```diff
@@ -3,6 +3,11 @@
 import Foundation
 import UIKit
 
+struct TransactionsSectionModel {
+    var header: String
+    var items: [Transaction]
```
vikmeup: This could be `let` both of them

vikmeup:
@OlegGordiichuk what will happen if you scroll down, you have new transaction added to the top section, when you scroll up - would it be there?

vikmeup:
@OlegGordiichuk is there any issues with getting new transactions? what's left on this PR?

vikmeup:
@OlegGordiichuk should we close this for now? seems like you have better solution

PR #318, opened 2018-02-13
show symbol for testnets

hewigovens:
Since we don't group testnets by network, it's clear to show symbol after testnet name.
e.g: Kovan -> Kovan (ETH)


hewigovens:
@vikmeup It's easy to distinguish production networks, only one mainnet.

vikmeup:
@hewigovens I would do it for clarify, for example we gonna be adding `Callisto` support, it would be `CLO`.

Let's do for both, make it clear for the user.

PR #320, opened 2018-02-13
Add handling for sync calls

vikmeup:


mishfit:
It would be closer to overriding the ProviderEngine

JS/src/index.js
```diff
@@ -56,5 +58,50 @@ if (typeof context.Trust === 'undefined') {
...
+ProviderEngine.prototype.send = function (payload) {
+  const self = this
+
+  let selectedAddress
+  let result = null
+  switch (payload.method) {
+
+    case 'eth_accounts':
+      // read from localStorage
+      selectedAddress = self.publicConfigStore.getState().selectedAddress
```
mishfit: Have to remove the publicConfigStore and other things that we don't have

Issue #123, opened 2017-12-10
Add Backup Private Key from wallets list


michaelScoff:
- Implement export private key functionality
- Add ability to export private key from info button in wallets list

yaleidev:
hi, are you solved?  @michaelScoff 
export private key?

michaelScoff:
@yaleidev yo! It’s not resolve it. Feel free to send pull request for it!

vikmeup:
https://github.com/TrustWallet/trust-wallet-ios/blob/master/Trust/EtherClient/EtherKeystore.swift

This is how we convert private key into keystore, if you do reverse operation you could get private key back

michaelScoff:
https://github.com/TrustWallet/trust-wallet-ios/blob/5609eadfd6db7708a15057f9b63ea8394594ffe4/Trust/EtherClient/EtherKeystore.swift#L267

This is how we convert private key into keystore, if you do reverse operation you could get private key back from the keystore

yaleidev:
func convertKeystoreFileToPrivateKey(value: String, passphrase: String) -> Result<String, KeystoreError> {
        
        guard let json = JSON.init(parseJSON: value).dictionaryObject else {
            return .failure(KeystoreError.failedToImportPrivateKey)
        }
        
        do {
            let passphraseBytes: [UInt8] = Array(passphrase.utf8)
            // reduce this number for higher speed. This is the default value, though. n
            
            let crypto = json["crypto"] as! [String: Any]
            
            let kdfparams = crypto["kdfparams"] as! [String: Any]
            let cipherparams = crypto["cipherparams"] as! [String: Any]
            let sa = kdfparams["salt"] as! String
            let numberOfIterations = kdfparams["n"] as! Int
            let ia = cipherparams["iv"] as! String
            let cip = crypto["ciphertext"] as! String
            
            Log(sa)
            Log(numberOfIterations)
            Log(ia)
            Log(cip)
            
            let salt: [UInt8] = Array(sa.utf8)
            
            let derivedKey = try PKCS5.PBKDF2(password: passphraseBytes, salt: salt, iterations: numberOfIterations, variant: .sha256).calculate()
            
            Log(derivedKey)
            
            let iv: [UInt8] = Array(ia.utf8)
            let aes = try AES(key: Array(derivedKey), blockMode: .CTR(iv: iv), padding: .noPadding)
            

            let ciphertext: [UInt8] = Array(cip.utf8)
            
[...]

PR #345, opened 2018-02-14
[WIP] Export Private Key

MillerApps:
Adding export private key functionality as mentioned in #123

### DIFF
--- a/Trust/EtherClient/EtherKeystore.swift
+++ b/Trust/EtherClient/EtherKeystore.swift
@@ -235,13 +235,26 @@ open class EtherKeystore: Keystore {
         guard let account = getAccount(for: account.address) else {
             return .failure(.accountNotFound)
         }
-
         do {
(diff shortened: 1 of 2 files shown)


Trust/EtherClient/EtherKeystore.swift
```diff
@@ -235,13 +235,26 @@ open class EtherKeystore: Keystore {
...
-
         do {
             let data = try keyStore.export(account: account, password: password, newPassword: newPassword)
             return (.success(data))
         } catch {
             return (.failure(.failedToDecryptKey))
         }
+
+    }
+    //TODO: Add error message
```
vikmeup: can you remove TODO? or implement it

I would return as `Data` and then you would convert to hex when needed.

MillerApps:
@MillerApps Okay will do, thanks for the feedback

PR #346, opened 2018-02-15
create signMessageData

James-Sangalli:
SignMessage will work for a pure string like "Hello" but fails when you create a message buffer of [UInt8] and cast it into a string using a utf8 format.

By taking the Data type directly you avoid this pitfall and make it easier on the developer as casting a buffer of [UInt8] works smoothly.


vikmeup:
@James-Sangalli please check master, I just implemented this: https://github.com/TrustWallet/trust-wallet-ios/pull/343

PR #347, opened 2018-02-15
Send max amount

hewigovens:
Change list:

1. Minus gas fee if user tries to send max balance of ethers.
2. Add "Total" row to confirm payment view if user send ethers
3. Disable the send button if no sufficient ethers
4. Add estimated max USD


OlegGordiichuk:
Hello @hewigovens could you pls move logic to view model of the send view controller? And add unit test for that?

OlegGordiichuk:
Try to test this part of the solution in different locale where is . and , Do sendMaxAmount() will return proper ;pea;ized value?

PR #351, opened 2018-02-15
Fix for Double value rounding.

OlegGordiichuk:


Trust/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -59,6 +59,14 @@ class SendViewController: FormViewController {
     lazy var stringFormatter: StringFormatter = {
         return StringFormatter()
     }()
+    lazy var decimals: Int = {
+        switch self.transferType {
+        case .ether:
+            return 8
```
vikmeup: why is ether decimal 8? `RPCServer.decimals` default is `18`

TrustTests/Foundation/StringFormatterTest.swift
```diff
@@ -8,5 +8,7 @@ class StringFormatterTest: XCTestCase {
     func testStringFormatter() {
         XCTAssertEqual(stringFormatter.formatter(for: 0.00001), "0.000010")
         XCTAssertEqual(stringFormatter.formatter(for: 0.1234, with: 2), "0.12")
+        XCTAssertEqual(stringFormatter.token(with: 0.1234, and: 1), "0.1")
+        XCTAssertEqual(stringFormatter.token(with: 1.1234, and: 0), "1")
```
vikmeup: can you add more test cases for large decimals as well?

Trust/Foundation/StringFormatter.swift
```diff
@@ -24,6 +32,17 @@ final class StringFormatter {
         formatter.currencyCode = currencyCode
         return formatter.string(from: NSNumber(value: value)) ?? "\(value)"
     }
+    /// Converts a Double to a `token String`.
+    ///
+    /// - Parameters:
+    ///   - double: double to convert.
+    ///   - decimals: symbols after coma.
+    /// - Returns: Token `String` represenation.
+    func token(with value: Double, and decimals: Int) -> String {
```
vikmeup: is it safe to use `Double` at all?we should be using string convert to formatter

Trust/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -322,14 +279,14 @@ extension SendViewController: UITextFieldDelegate {
...
         }
-        let text = (input as NSString).replacingCharacters(in: range, with: string)
-        guard let amount = decimalFormatter.number(from: text) else {
-            //Should be done in another way.
-            pairValue = 0.0
-            updatePriceSection()
-            return true
-        }
-        self.updatePairPrice(with: amount.doubleValue)
+        //Total amount of the user input.
```
vikmeup: can we move this login into view model?

Trust/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -240,22 +204,26 @@ class SendViewController: FormViewController {
...
+    private func updatePairPrice(with amount: Decimal) {
+        guard let rates = storage.tickers, let currentTokenInfo = rates[viewModel.destinationAddress.description], let price = Decimal(string: currentTokenInfo.price) else {
             return
         }
-        if self.currentPair.left == viewModel.symbol {
-            pairValue = amount * price
+        viewModel.updatePaitRate(with: price, and: amount)
+        self.updatePriceSection()
+    }
+    private func updateAmount(with stringAmount: String) {
```
vikmeup: can we live in the view model as well?

PR #356, opened 2018-02-18
Avoid showing pin on will resign active

vikmeup:


OlegGordiichuk:
Run on iPhone 6 seams to me working fine. Need more testing with face id.

vikmeup:
I will cover Face ID today

PR #357, opened 2018-02-18
Update to latest version of Keystore

aleph7:


vikmeup:
Can you take a look at failing test? Then good to go 💪

imzxin:
Type 'BlockMode' has no member 'CTR' , ype 'BlockMode' has no member 'CBC

Issue #315, opened 2018-02-12
Disable send button if you don't have enough funds to pay for gas fee


vikmeup:
It should be disable on confirmation view page and it should say `no enough ether` or something similar

Issue #33, opened 2017-10-03
Support send max amount 


michaelScoff:

colourful-land:
It certainly seem to be tricky to implement thanks to Ethere's fee-outside-transaction model different to Bitcoin's. The wallet I use (myetherwallet) requires two attempts to sweep an account - it first guesses a gas value, then, failing a transaction, ask me to sign a new one, and this time it would be able to empty the entire balance. That gives a hint that this is not strait-forward to do.

michaelScoff:
@hewigovens can you take on this?

hewigovens:
👌

colourful-land:
@hewigovens Can you hint how this should be done? The gas is unpredictable, do you plan to send a transaction with the intention to fail out of gas, then update the amount and try again?

michaelScoff:
@colourful-land you can check if your transaction will fail before sending. We should not let users send funds if it's intended to fail

hewigovens:
@colourful-land we will update the UI with amount and fee first, then we will try to check if the transaction will fail before sending.

PR #360, opened 2018-02-20
Support send max amount ether / token

hewigovens:
For #33 and #315 

1. Minus gas fee if user tries to send max balance.
2. Add "Total" row to confirm payment view if user send ether
3. Disable the send button if no sufficient ether
4. Fix updatePaitRate typo

### DIFF
--- a/Trust/Localization/en.lproj/Localizable.strings
+++ b/Trust/Localization/en.lproj/Localizable.strings
@@ -48,6 +48,7 @@
 "send.error.emptyClipBoard" = "Empty ClipBoard";
 "send.error.invalidAddress" = "Invalid Address";
 "send.error.wrongInput" = "Wrong Input";
+"send.error.insufficientFund" = "Insufficient Ethers";
 "send.max.button.title" = "Max";
 "send.paste.button.title" = "Paste";
 "send.recipientAddress.textField.placeholder" = "Recipient Address";
--- a/Trust/Localization/en.lproj/Localizable.strings
+++ b/Trust/Localization/en.lproj/Localizable.strings
@@ -48,7 +48,7 @@
 "send.error.emptyClipBoard" = "Empty ClipBoard";
 "send.error.invalidAddress" = "Invalid Address";
 "send.error.wrongInput" = "Wrong Input";
-"send.error.insufficientFund" = "Insufficient Ethers";
+"send.error.insufficientFund" = "Insufficient Fund";
 "send.max.button.title" = "Max";
 "send.paste.button.title" = "Paste";
 "send.recipientAddress.textField.placeholder" = "Recipient Address";
--- a/Trust/Transfer/Controllers/TransactionConfigurator.swift
+++ b/Trust/Transfer/Controllers/TransactionConfigurator.swift
@@ -192,22 +192,16 @@ class TransactionConfigurator {
         var gasValue = transaction.gasPrice * transaction.gasLimit
         //Amount of the tokens.
         var tokenValue = BigInt(0)
-        var tokenObject: TokenObject?
         //We check if it is ETH or token operation.
         switch transaction.transferType {
         case .ether:
             gasValue += transaction.value
+            return gasValue <= balance.value
         case .token(let token):
-            tokenObject = token
             tokenValue += transaction.value
+            let isEnoughOfEther = gasValue <= balance.value
+            let isEnoughOfTokens = tokenValue <= token.valueBigInt
+            return isEnoughOfTokens && isEnoughOfEther
         }
-        let isEnoughOfEther = gasValue <= balance.value
-        //If it is ETH transaction validate only isEnoughOfEther
-        guard let token = tokenObject else {
-            return isEnoughOfEther
-        }
-        let isEnoughOfTokens = tokenValue <= token.valueBigInt
-        //If it is token operation validate token amount and eth for transaction fees.
-        return isEnoughOfTokens && isEnoughOfEther
     }
 }
--- a/Trust/Transfer/Controllers/TransactionConfigurator.swift
+++ b/Trust/Transfer/Controllers/TransactionConfigurator.swift
@@ -129,9 +129,19 @@ class TransactionConfigurator {
         }
     }
 
+    func valueToSend() -> BigInt {
+        var value = transaction.value
+        if let balance = session.balance?.value,
+            balance == transaction.value {
+            // user tries to send max
+            value = transaction.value - configuration.gasLimit * configuration.gasPrice
+        }
+        return value
+    }
+
     func previewTransaction() -> PreviewTransaction {
         return PreviewTransaction(
-            value: transaction.value,
+            value: self.valueToSend(),
             account: account,
             address: transaction.to,
             contract: .none,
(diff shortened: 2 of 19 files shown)


vikmeup:
@OlegGordiichuk can you take a look at this?

OlegGordiichuk:
@vikmeup  and @hewigovens .

In this commit I have done this staff:

- I rewrite logic that is in sendMaxAmount() function. 
- Add new function to EtherNumberFormatter to return Decimal.
- Add new function to StringFormatter to return Decimal with respect of the locale decimal separator.
- Add Unit tests.

For approximate calculation of the FIAT or Crypto value we are using  base-ten positional numeral system in our case it is Decimal or BigInt.So for more easier conversion of the rates we are using decimals in send view controller.

Problem was in testing of the Decimal that is presented as String because of the locale decimal separator. 

**You try to compare "1,23"  with  "1.23".**

But you can use the simplest way:

**let decimalSeparator = Locale.current.decimalSeparator ?? "."

"1\(decimalSeparator)23"  with  "1.23".**

In this way it will be locale agnostic.

Trust/Transfer/ViewControllers/ConfirmPaymentViewController.swift
```diff
@@ -132,6 +141,26 @@ class ConfirmPaymentViewController: UIViewController {
...
         displayChildViewController(viewController: stackViewController)
+        updateSubmitButton()
+    }
+
+    private func updateSubmitButton() {
+        guard let balance = session.balance else {
+            return
+        }
+
+        let transaction = configurator.previewTransaction()
```
vikmeup: can we move this logic into `viewModel`?

Trust/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -35,6 +35,7 @@ class SendViewController: FormViewController {
     var amountRow: TextFloatLabelRow? {
         return form.rowBy(tag: Values.amount) as? TextFloatLabelRow
     }
+    var maxButton: UIButton?
```
vikmeup: can you make `maxButton` lazy initializer? I would prefer to avoid optional variables

OlegGordiichuk: Done.

Trust/Transfer/ViewControllers/SendViewController.swift
```diff
@@ -183,7 +184,7 @@ class SendViewController: FormViewController {
         //Update button title.
         sender.setTitle(viewModel.currentPair.right, for: .normal)
         //Hide max button
-        maxButton?.isHidden = viewModel.currentPair.left != viewModel.symbol
+        maxButton.isHidden = viewModel.currentPair.left != viewModel.symbol
```
vikmeup: can you move this into view model as well. `isMaxButtonHidden: Bool`?

OlegGordiichuk: Done and add unit test.

Trust/Localization/en.lproj/Localizable.strings
```diff
@@ -48,7 +48,7 @@
 "send.error.emptyClipBoard" = "Empty ClipBoard";
 "send.error.invalidAddress" = "Invalid Address";
 "send.error.wrongInput" = "Wrong Input";
-"send.error.insufficientFund" = "Insufficient Ethers";
+"send.error.insufficientFund" = "Insufficient Fund";
```
vikmeup: don't edit here, it will be done on localize

Trust/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -183,4 +183,31 @@ class TransactionConfigurator {
     func update(configuration: TransactionConfiguration) {
         self.configuration = configuration
     }
+    func isValidBalance() -> Bool {
+        guard let balance = session.balance else {
+            return false
+        }
+        let transaction = previewTransaction()
+        //Amount of the gas that is required for transaction.
+        var ethValue = transaction.gasPrice * transaction.gasLimit
```
vikmeup: `ethValue` = > `gasValue`

Trust/Transfer/ViewControllers/ConfirmPaymentViewController.swift
```diff
@@ -145,22 +145,9 @@ class ConfirmPaymentViewController: UIViewController {
...
-        }
-        if value > balance.value {
-            submitButton.isEnabled = false
-            submitButton.setTitle(viewModel.insufficientFundText, for: .normal)
-        } else {
-            submitButton.isEnabled = true
-            submitButton.setTitle(viewModel.actionButtonText, for: .normal)
-        }
+        submitButton.isEnabled = configurator.isValidBalance()
+        let text = configurator.isValidBalance() ? viewModel.actionButtonText : viewModel.insufficientFundText
```
vikmeup: `configurator.isValidBalance() ? viewModel.actionButtonText : viewModel.insufficientFundText`

this should be moved into view model

Trust/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -183,4 +183,31 @@ class TransactionConfigurator {
...
+        //Amount of the tokens.
+        var tokenValue = BigInt(0)
+        var tokenObject: TokenObject?
```
vikmeup: can you avoid option variable?

Trust/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -183,4 +183,31 @@ class TransactionConfigurator {
...
+        //We check if it is ETH or token operation.
+        switch transaction.transferType {
+        case .ether:
+            ethValue += transaction.value
+        case .token(let token):
+            tokenObject = token
+            tokenValue += transaction.value
+        }
+        let isEnoughOfEther = ethValue <= balance.value
```
vikmeup: If balance endpoint if down you won't be able to send anything right? that would suck

vikmeup:
@hewigovens @OlegGordiichuk can you guys take a final look?

Trust/Transfer/Controllers/TransactionConfigurator.swift
hewigovens: We can add another func like isTokenBalanceValid(_ token: TokenObject)

vikmeup: I like that

OlegGordiichuk: I optimize this function. so there is no need to add additional one .

hewigovens:
👍

Trust/Transfer/Controllers/TransactionConfigurator.swift
```diff
@@ -192,22 +192,16 @@ class TransactionConfigurator {
...
         var tokenValue = BigInt(0)
-        var tokenObject: TokenObject?
         //We check if it is ETH or token operation.
         switch transaction.transferType {
         case .ether:
             gasValue += transaction.value
+            return gasValue <= balance.value
         case .token(let token):
-            tokenObject = token
             tokenValue += transaction.value
```
vikmeup: `tokenValue` only being used once, you can move it into token case

PR #361, opened 2018-02-20
Fix #321

### LINKED ISSUE #321: Update URL in the browser
If you are using react native applications and press on different links it won't change URL in the browser.

Try: https://rarebits.io/item/CryptoKitties/366900 and then typing on any of the links on the website, it won't change URL in the textfield

hewigovens:
reference: https://stackoverflow.com/questions/47829963/how-can-i-detect-when-url-of-amp-page-changed-with-wkwebview


vikmeup:
@hewigovens this is sooooooo great! it's been annoying 

can you remove `refreshURL()`, that's not necessary right?

hewigovens:
I'm testing without refreshURL()

PR #382, opened 2018-02-21
Fix Bug 379

hewigovens:


Trust/Wallet/ViewControllers/TrustDocumentPickerViewController.swift
```diff
@@ -0,0 +1,16 @@
+// Copyright SIX DAY LLC. All rights reserved.
+
+import UIKit
+
+class TrustDocumentPickerViewController: UIDocumentPickerViewController {
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        UINavigationBar.appearance().tintColor = Colors.blue
```
vikmeup: Should we move this into `AppStyle` as well?

hewigovens: Sure, which variable should we use?

PR #399, opened 2018-02-24
Remove BonMot lib

MillerApps:
Removed BonMot lib as mentioned in issue #389


vikmeup:
@MillerApps nice work 👍

PR #400, opened 2018-02-24
Add BalanceStatus for insufficient wordings

hewigovens:
Add wordings for these cases:

- insufficientEther
- insufficientGas
- insufficientToken

### DIFF
--- a/Trust.xcodeproj/project.pbxproj
+++ b/Trust.xcodeproj/project.pbxproj
@@ -359,6 +359,7 @@
 		A2B7ECAC2038FFBC00828E7D /* ExportPrivateKeyViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = A2B7ECAB2038FFBC00828E7D /* ExportPrivateKeyViewModel.swift */; };
 		A2B7ECAE2039001300828E7D /* ExportPrivateKeyViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = A2B7ECAD2039001300828E7D /* ExportPrivateKeyViewController.swift */; };
 		BB03A86D20416AF600649DA0 /* BalanceStatus.swift in Sources */ = {isa = PBXBuildFile; fileRef = BB03A86C20416AF600649DA0 /* BalanceStatus.swift */; };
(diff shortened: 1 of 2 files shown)


Trust/Transfer/Types/BalanceStatus.swift
```diff
@@ -0,0 +1,76 @@
...
+            if !gasSufficient {
+                return .insufficientGas
+            }
+        }
+        return .none
+    }
+
+    var insufficientText: String {
+        let key = insufficientTextKey
+        if key == Key.insufficientEther {
```
vikmeup: can we do switch statement in here instead?

Trust/Transfer/Types/BalanceStatus.swift
```diff
@@ -0,0 +1,76 @@
...
+                return "send.error.insufficientGas"
+            case .insufficientToken:
+                return "send.error.insufficientToken"
+            case .none:
+                return ""
+            }
+        }
+    }
+
+    private func _L(_ key: String, value: String? = .none) -> String {
```
vikmeup: Can we avoid `_L` method? make it hard to read. Using `NSLocalizedString ` should be ok

Trust/Transfer/Types/BalanceStatus.swift
```diff
@@ -0,0 +1,76 @@
...
+        return NSLocalizedString(key, value: value ?? key, comment: "")
+    }
+
+    var sufficient: Bool {
```
vikmeup: I would have separate test unit test just for `BalanceStatus` as well

vikmeup:
@hewigovens Did you consider having `BalanceStatus` as keys itself? 
```
enum BalanceStatus {
     case ether(etherSufficient: Bool, gasSufficient: Bool)
     case token(tokenSufficient: Bool, gasSufficient: Bool)
 }
```
=> 
```
enum BalanceStatus {
    case insufficientEther
    case insufficientGas
    case insufficientToken
    case correct
}
```

hewigovens:
@vikmeup If we make BalanceStatus flat, it will lose the context for insufficientGas

OlegGordiichuk:
Nice work.

PR #408, opened 2018-02-26
Private Key added to memory when Button held

MillerApps:
Made the private Key and QR Code generate upon holding the button.

### DIFF
--- a/Trust/Export/ViewControllers/ExportPrivateKeyViewController.swift
+++ b/Trust/Export/ViewControllers/ExportPrivateKeyViewController.swift
@@ -90,8 +90,10 @@ class ExportPrivateKeyViewConroller: UIViewController {
             revalQRCodeButton.trailingAnchor.constraint(equalTo: stackView.layoutMarginsGuide.trailingAnchor),
             revalQRCodeButton.leadingAnchor.constraint(equalTo: stackView.layoutMarginsGuide.leadingAnchor),
 
-            imageView.widthAnchor.constraint(lessThanOrEqualToConstant: Layout.widthAndHeight),
             imageView.heightAnchor.constraint(equalToConstant: Layout.widthAndHeight),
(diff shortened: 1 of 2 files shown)


Trust/Export/ViewControllers/ExportPrivateKeyViewController.swift
```diff
@@ -93,21 +93,21 @@ class ExportPrivateKeyViewConroller: UIViewController {
...
-        createQRCode()
     }
 
     func blur(image: UIImageView) {
-        let blur = UIBlurEffect(style: .extraLight)
+        let blur = UIBlurEffect(style: .dark)
         let view = UIVisualEffectView(effect: blur)
 
-        view.frame = image.bounds
+        view.frame = CGRect(x: -130, y: -5, width: Layout.widthAndHeight, height: 10) // Needed to postion blur in corrct space, setting to image.bounds = 0,0,0,0, thus the postion is incorrect.
```
vikmeup: Do not hardcode any pixel values.

MillerApps: Fixed using auto layout

vikmeup:
@MillerApps did you look into different wallets, how do they do private key reveal?

MillerApps:
@vikmeup I didn’t look at any iOS wallets, I did look at MEW and Metamask though. MEW shows a blacked area with a clickable bit of text that says show which reveals a QR Code, they also have an option to copy the private key as text. Metamask has an option to copy the private key as text.

MillerApps:
@vikmeup Your thoughts?

vikmeup:
Should be ok for now, I don't really like show this view looks like tho. I will refine this later.

MillerApps:
Sounds good, I'm not not a fan of the UI for it either but, I couldn't think of any alternatives.

PR #410, opened 2018-02-27
Show symbol in insufficient wordings

### LINKED ISSUE #407: Display insufficient balance
On transaction confirmation page `insufficient ethers` should reflect network symbol of current network:
Example for POA: `insufficient POA balance`
Example for ETH: `insufficient ETH balance`

hewigovens:
Try to fix #407

### DIFF
--- a/Trust/Transfer/ViewControllers/ConfirmPaymentViewController.swift
+++ b/Trust/Transfer/ViewControllers/ConfirmPaymentViewController.swift
@@ -147,8 +147,11 @@ class ConfirmPaymentViewController: UIViewController {
     private func updateSubmitButton() {
         let status = configurator.balanceValidStatus()
         submitButton.isEnabled = status.sufficient
-        let text = status.sufficient ? viewModel.actionButtonText : status.insufficientText
-        submitButton.setTitle(text, for: .normal)
+        if status.sufficient {
+            submitButton.setTitle(viewModel.actionButtonText, for: .normal)
+        } else {
+            submitButton.setTitle(viewModel.getInsufficientText(status, config: configurator.session.config, transferType: configurator.transaction.transferType), for: .normal)
+        }
     }
 
     private func reloadView() {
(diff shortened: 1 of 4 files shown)


Trust/Transfer/ViewModels/ConfirmPaymentViewModel.swift
```diff
@@ -27,4 +27,29 @@ struct ConfirmPaymentViewModel {
...
+        case .token(let token):
+            switch status {
+            case .token(let tokenSufficient, let gasSufficient):
+                if !tokenSufficient {
+                    return String(format: format, token.symbol)
+                }
+                if !gasSufficient {
+                    return String(format: format, networkSymbol)
+                }
+                // should not be here
```
vikmeup: can you return `.none` instead? Then it's clear

Trust/Transfer/ViewControllers/ConfirmPaymentViewController.swift
```diff
@@ -147,8 +147,11 @@ class ConfirmPaymentViewController: UIViewController {
     private func updateSubmitButton() {
         let status = configurator.balanceValidStatus()
         submitButton.isEnabled = status.sufficient
-        let text = status.sufficient ? viewModel.actionButtonText : status.insufficientText
-        submitButton.setTitle(text, for: .normal)
+        if status.sufficient {
```
vikmeup: can you do `submitButton.setTitle(viewModel.actionButtonText, for: .normal)` ? and put all the login into view model?

Trust/Transfer/ViewModels/ConfirmPaymentViewModel.swift
hewigovens: but the type is not optional?

vikmeup: If we have such cases, we need to review design architecture for this.

It might be ok for now, but it would be good to refactor this part

hewigovens: Agreed, Let's discuss better solution offline.

hewigovens:
We'll refactor this part in later version.

PR #419, opened 2018-03-03
Implement HD Wallet

vikmeup:
What do you think about this? HD Wallet and Regular Wallet? 

simulator screen shot - iphone x - 2018-03-02 at 20 27 40

### DIFF
--- a/Trust/Accounts/ViewModels/AccountsViewModel.swift
+++ b/Trust/Accounts/ViewModels/AccountsViewModel.swift
@@ -5,13 +5,67 @@ import TrustKeystore
 
 struct AccountsViewModel {
 
-    let wallets: [Wallet]
+    private let regularWallets: [Wallet]
+    private let hdWallets: [Wallet]
 
     init(wallets: [Wallet]) {
-        self.wallets = wallets
+        self.hdWallets = wallets.filter { wallet in
+            switch wallet.type {
+            case .real(let account):
+                return account.type == .hierarchicalDeterministicWallet
+            case .watch:
+                return false
+            }
+        }
+        self.regularWallets = wallets.filter { wallet in
+            switch wallet.type {
+            case .real(let account):
+                return account.type == .encryptedKey
+            case .watch:
+                return true
+            }
+        }
     }
 
     var title: String {
         return NSLocalizedString("wallet.navigation.title", value: "Wallets", comment: "")
+    }
+
+    var numberOfSections: Int {
+        return 2
+    }
+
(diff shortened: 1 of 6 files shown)


Trust/Accounts/ViewModels/AccountsViewModel.swift
```diff
@@ -5,13 +5,67 @@ import TrustKeystore
 
 struct AccountsViewModel {
 
-    let wallets: [Wallet]
+    private let regularWallets: [Wallet]
+    private let hdWallets: [Wallet]
 
     init(wallets: [Wallet]) {
-        self.wallets = wallets
+        self.hdWallets = wallets.filter { wallet in
```
vikmeup: @aleph7 what's the better way to split wallet in here? Looks little tricky

MillerApps:
I think the UI looks fine.

vikmeup:
@MillerApps I don't think it's a UI question here. UX wise HD Wallet and Regular wallet could be confusing names :(

MillerApps:
@vikmeup excellent point, I’m sure no matter you do some will still be confused. I wish I had better feedback for you. 

aleph7:
What about HD Wallets and Encrypted Keys

Trust/Accounts/ViewModels/AccountsViewModel.swift
aleph7: What about instead of having `real` and `watch` we have `hd`, `privateKey`, and `address` for the wallet type.

aleph7: We could move this to keystone as well.

vikmeup:
`Regular Wallet` or `Encrypted Keys` might too confusing to the user. I'm thinking to remove title for "regular wallet" at all

vikmeup:
@alejandro-isaza-hs I'm also wondering about a derived path, should we have that option? since different network could have a different derived path?

aleph7:
Yes, maybe have it as an advanced option.

vikmeup:
@aleph7 I assume current implementation works with Meta Mask right? how about Jaxx?

aleph7:
Not sure what you mean. How are they different?

vikmeup:
@aleph7 some wallets have their own derived path, like Coinomi app. I would need to check if they all have the same derived path. 

I'm going to merge this for now and then add an advanced option for future releases.

aleph7:
Yes, if you can manually enter the path then we support them all.

PR #420, opened 2018-03-03
Add trasnaction hash

### LINKED ISSUE #418: Return transaction ID when transaction signed

aleph7:
Fix #418


aleph7:
Is this what you need @vikmeup ?

vikmeup:
Thanks! That should work I think.

PR #422, opened 2018-03-04
Implement bookmarks for Dapp browser

VKoskiv:
This isn't fully done yet. Looking for some feedback on architecture.


Trust/Browser/Storage/BookmarksStore.swift
```diff
@@ -0,0 +1,26 @@
...
+    var bookmarks: Results<BookmarkObject> {
+        return realm.objects(BookmarkObject.self)
+    }
+    let realm: Realm
+    init(
+        realm: Realm
+    ) {
+        self.realm = realm
+    }
+    func add(bookmarks: [BookmarkObject]) {
```
vikmeup: @OlegGordiichuk duplication code in here, what's the best way to combine it?

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,98 @@
+// Copyright SIX DAY LLC. All rights reserved.
+
+import Foundation
+import UIKit
+
+protocol BookmarkViewControllerDelegate: class {
+    func didSelectBookmark(bookmark: BookmarkObject, in viewController: BookmarkViewController)
```
vikmeup: I would suggest to call:
```
func didSelectBookmark(_ bookmark: BookmarkObject, in viewController: BookmarkViewController)
```
or

``````
func didSelect(bookmark: BookmarkObject, in viewController: BookmarkViewController)
```

I believe the first option is easier.

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+    func didDeleteBookmark(bookmark: BookmarkObject, in viewController: BookmarkViewController)
+}
+
+class BookmarkViewController: UITableViewController {
+    weak var delegate: BookmarkViewControllerDelegate?
+    var allowsBookmarkDeletion: Bool = true
```
vikmeup: `allowsBookmarkDeletion`, not necessary atm. Let's remove it until we need it.

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+    var headerTitle: String?
```
vikmeup: `headerTitle` let's remove it

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+            confirmDelete(bookmark: bookmark)
+        }
+    }
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let bookmark = self.bookmark(for: indexPath)
+        delegate?.didSelectBookmark(bookmark: bookmark, in: self)
+    }
+    func confirmDelete(bookmark: BookmarkObject) {
+        confirm(title: NSLocalizedString("browser.bookmarks.confirm.delete.title", value: "Are you sure you would like to delete this bookmark?", comment: ""),
+                okTitle: NSLocalizedString("browser.bookmarks.confirm.delete.okTitle", value: "Delete", comment: ""),
```
vikmeup: `Delete` is a generic word, use key as `Delete`

Trust/Browser/ViewControllers/BrowserViewController.swift
```diff
@@ -196,6 +206,16 @@ class BrowserViewController: UIViewController {
         navigationController.present(controller, animated: true, completion: nil)
     }
 
+    private func addBookmark() {
+        print("Add Bookmark pressed")
+        guard let url = webView.url else { return }
+        //Save this to user bookmarks
```
vikmeup: Simplest would be saving to the store, without any confirmation.

vikmeup: + title

Trust/Browser/Coordinators/BookmarkCoordinator.swift
```diff
@@ -0,0 +1,57 @@
+// Copyright SIX DAY LLC. All rights reserved.
+
+import Foundation
+import UIKit
+
+protocol BookmarksCoordinatorDelegate: class {
+    func didCancel(in coordinator: BookmarksCoordinator)
+    func didSelectBookmark(bookmark: BookmarkObject, in coordinator: BookmarksCoordinator)
```
vikmeup: also let's rename `BookmarkObject` => `Bookmark`

vikmeup:
@VKoskiv how's progress on this?

VKoskiv:
It's working as expected on the bookmarks side, but there's a new bug.
When loading up different bookmarks, after a few page loads, the `func goTo(url: URL)` in BrowserViewController.swift:98 appears to stop working.
It is called properly, and the page load bar is shown, but the page never changes.

Trust/Browser/Coordinators/BrowserCoordinator.swift
```diff
@@ -80,6 +84,21 @@ class BrowserCoordinator: Coordinator {
...
+            store: bookmarksStore
+        )
+        coordinator.delegate = self
+        coordinator.start()
+        addCoordinator(coordinator)
+        navigationController.present(coordinator.navigationController, animated: true, completion: nil)
+    }
+    
+    func openBookmark(bookmark: Bookmark) {
+        rootViewController.goTo(url: URL(string: bookmark.url)!)
```
VKoskiv: This works as expected on the bookmark side of things, but after a few page loads on different bookmarks, `goTo()` appears to stop working.
It is called with the proper URL, and it shows the page loading bar, but the page content never changes.

VKoskiv: Can't seem to reproduce it now 😅

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+                    switch result {
+                    case .success:
+                        self.delete(bookmark: bookmark, index: index)
+                    case .failure: break
+                    }
+        }
+    }
+    func delete(bookmark: Bookmark, index: IndexPath) {
+        store.delete(bookmarks: [bookmark])
+        //FIXME: How do we add an animation to this? Can't do it manually since remove() already removes it
```
VKoskiv: The `self.bookmarks.remove()` triggers the `tableView.reloadData()` in the `didSet` of the `bookmarks` array.
Removing the row manually with a fade effect, something like:
`tableView.deleteRows(at: [indexPath], with: .fade)`
Can't be done here.

Trust/Browser/Coordinators/BrowserCoordinator.swift
vikmeup: can you avoid force unwrapping?

Trust/Browser/ViewControllers/BookmarkViewController.swift
vikmeup: No animation is good 👍

Trust/Browser/Types/Bookmark.swift
```diff
@@ -5,11 +5,25 @@ import RealmSwift
...
+        self.id = incrementID()
+    }
+
+    override class func primaryKey() -> String? {
+        return "id"
+    }
+
+    func incrementID() -> Int {
+        let realm = try! Realm()
+        return (realm.objects(Bookmark.self).max(ofProperty: "id") as Int? ?? 0) + 1
```
vikmeup: I would default `0`=> `1`

Trust/Browser/Coordinators/BookmarkCoordinator.swift
```diff
@@ -6,52 +6,36 @@ import UIKit
 protocol BookmarksCoordinatorDelegate: class {
     func didCancel(in coordinator: BookmarksCoordinator)
     func didSelectBookmark(bookmark: Bookmark, in coordinator: BookmarksCoordinator)
-    func didAddBookmark(bookmark: Bookmark, in coordinator: BookmarksCoordinator)
-    func didDeleteBookmark(bookmark: Bookmark, in coordinator: BookmarksCoordinator)
 }
 
 class BookmarksCoordinator: Coordinator {
```
vikmeup: Can you add basic unit tests for `BookmarkCoordinator` to make sure `start` will show correct view controller

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+    weak var delegate: BookmarkViewControllerDelegate?
+    var headerTitle: String?
+    var viewModel: BookmarkViewModel {
+        return BookmarkViewModel(
+            bookmarks: bookmarks
+        )
+    }
+    var hasBookmarks: Bool {
+        return !store.bookmarks.isEmpty
+    }
```
vikmeup: Can you move `hasBookmarks` into view model?

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+        return viewModel.bookmarks[indexPath.row]
+    }
+    override func numberOfSections(in tableView: UITableView) -> Int {
+        return 1
+    }
+    override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        return viewModel.bookmarks.count
+    }
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        //let cell = self.tableView.dequeueReusableCell(withIdentifier: "newCell", for: indexPath)
```
vikmeup: Can we use `dequeueReusableCell`?

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+        let cell = UITableViewCell()
+        cell.textLabel?.text = viewModel.bookmarks[indexPath.row].title
```
vikmeup: can we add url to subtitle?

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+            confirmDelete(bookmark: bookmark, index: indexPath)
+        }
+    }
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let bookmark = self.bookmark(for: indexPath)
+        delegate?.didSelectBookmark(bookmark: bookmark, in: self)
+    }
+    func confirmDelete(bookmark: Bookmark, index: IndexPath) {
+        confirm(title: NSLocalizedString("browser.bookmarks.confirm.delete.title", value: "Are you sure you would like to delete this bookmark?", comment: ""),
+                okTitle: NSLocalizedString("browser.bookmarks.confirm.delete.okTitle", value: "Delete", comment: ""),
```
vikmeup: Just use default key: `NSLocalizedString("Delete", value: "Delete", comment: "")`

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+                    case .success:
+                        self.delete(bookmark: bookmark, index: index)
+                    case .failure: break
+                    }
+        }
+    }
+    func delete(bookmark: Bookmark, index: IndexPath) {
+        store.delete(bookmarks: [bookmark])
+        //FIXME: How do we add an animation to this? Can't do it manually since remove() already removes it
+        self.bookmarks.remove(at: index.row)
```
vikmeup: Since you need to call two functions to delete, can you make it where you would delete from the store and then call fetch to refresh latest?

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+    func fetch() {
+        bookmarks = Array(store.bookmarks)
+    }
+    func configure(viewModel: BookmarkViewModel) {
+        title = headerTitle ?? viewModel.title
+    }
+    func bookmark(for indexPath: IndexPath) -> Bookmark {
+        return viewModel.bookmarks[indexPath.row]
+    }
+    override func numberOfSections(in tableView: UITableView) -> Int {
```
vikmeup: can you remove `numberOfSections`? I think default is 1

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        tableView.rowHeight = UITableViewAutomaticDimension
+        tableView.estimatedRowHeight = 55
+    }
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        fetch()
+    }
+    func fetch() {
```
vikmeup: can you leave spacing between method?

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -0,0 +1,96 @@
+// Copyright SIX DAY LLC. All rights reserved.
+
+import Foundation
+import UIKit
+
+protocol BookmarkViewControllerDelegate: class {
+    func didSelectBookmark(bookmark: Bookmark, in viewController: BookmarkViewController)
```
vikmeup: do `bookmark: Bookmark` => `_ bookmark: Bookmark`

Trust/Browser/ViewControllers/BookmarkViewController.swift
VKoskiv: I don't think so? This view doesn't use `R` for the TableViewCell view model.

vikmeup:
@VKoskiv let's make those small changes and 🚢  it!

Trust/Browser/ViewControllers/BookmarkViewController.swift
vikmeup: I think you can register `UITableViewCell` as a class for this tableView and reuse then

VKoskiv:
There is a fairly trivial way to fetch `favicons` for these bookmarks. Perhaps next version? [https://stackoverflow.com/questions/19424654/uiwebview-get-favicons](StackOverflow)

vikmeup:
@VKoskiv would this work? this might be simpler. I assume `favicon.ico` located in the root directory of the host

```swift
imageView.kf.setImage(
            with: host / favicon.ico,
            placeholder: R.image.launch_screen_logo()
        )
```

VKoskiv:
That could work. Do you want to include it in this build?

vikmeup:
@VKoskiv that sounds like an easy feature to add? if it works.

VKoskiv:
Checking right now. In `BrowserViewController` I just have access to the `webView`

VKoskiv:
Here's how Firefox does it: [https://github.com/mozilla-mobile/firefox-ios/blob/master/Client/Frontend/Browser/FaviconManager.swift](url)

vikmeup:
Try this
`import import Kingfisher`
```swift
imageView.kf.setImage(
            with: "https://www.google.com/favicon.ico",
            placeholder: R.image.launch_screen_logo()
        )
```

Simple call for each cell

VKoskiv:
There doesn't appear to be a `setImage()` function in `kf`. I assume `imageView` is just an `UIImageView`?

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -63,6 +64,11 @@ class BookmarkViewController: UITableViewController {
         let cell = UITableViewCell(style: .subtitle, reuseIdentifier: "newCell")
         cell.textLabel?.text = viewModel.bookmarks[indexPath.row].title
         cell.detailTextLabel?.text = viewModel.bookmarks[indexPath.row].url
+        cell.imageView?.kf.setImage(
+            with: URL(string: "\(viewModel.bookmarks[indexPath.row].url)favicon.ico"),
+            placeholder: R.image.launch_screen_logo())
```
vikmeup: can you move URL into view model?

vikmeup:
@VKoskiv what's left for this to be merged?

VKoskiv:
It's complete now, unless you have more suggestions.

Trust/Browser/Views/BookmarkViewCell.swift
```diff
@@ -0,0 +1,19 @@
+// Copyright SIX DAY LLC. All rights reserved.
+
+import Foundation
+import UIKit
+
+class BookmarkViewCell: UITableViewCell {
+    @IBOutlet weak var TitleLabel: UILabel!
```
vikmeup: can you lowecase variables? 
Pattern would be this:
```
titleLabel: UILabel, but since titleLabel exist in `UITableViewCell` then bookmarkTitleLabel: UILabel
```

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -60,15 +61,13 @@ class BookmarkViewController: UITableViewController {
...
-    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = UITableViewCell(style: .subtitle, reuseIdentifier: "newCell")
-        cell.textLabel?.text = viewModel.bookmarks[indexPath.row].title
-        cell.detailTextLabel?.text = viewModel.bookmarks[indexPath.row].url
-        cell.imageView?.kf.setImage(
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> BookmarkViewCell {
+        let cell = self.tableView.dequeueReusableCell(withIdentifier: R.nib.bookmarkViewCell.name, for: indexPath) as! BookmarkViewCell
+        cell.viewModel = BookmarkViewModel(bookmark: viewModel.bookmarks[indexPath.row])
+        cell.FaviconImage?.kf.setImage(
             with: URL(string: "\(viewModel.bookmarks[indexPath.row].url)favicon.ico"),
```
vikmeup:
```
URL(string: "\(viewModel.bookmarks[indexPath.row].url)favicon.ico")
```

move this into view model as `imageURL: URL?`, controller does not need to do all the logic

vikmeup: This is great for view model testing as well

Trust/Browser/ViewModel/BookmarkViewModel.swift
```diff
@@ -3,16 +3,16 @@
...
-
-    var hasBookmarks: Bool {
-        return !bookmarks.isEmpty
+    var url: String {
+        return self.bookmark.url
     }
-
     var title: String {
-        return NSLocalizedString("browser.bookmarks.title", value: "Bookmarks", comment: "")
+        return self.bookmark.title
```
vikmeup: Suppress `self.` it's not necessary here. 

You need to use `self.` in the case of `init()` function of if used in the closure

Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -37,6 +37,7 @@ class BookmarkViewController: UITableViewController {
         super.viewDidLoad()
         tableView.rowHeight = UITableViewAutomaticDimension
         tableView.estimatedRowHeight = 55
+        tableView.register(R.nib.bookmarkViewCell(), forCellReuseIdentifier: R.nib.bookmarkViewCell.name)
```
vikmeup: 👍

vikmeup:
@VKoskiv Nice work! this is a big improvement to dapp browser

PR #426, opened 2018-03-05
Issues/557

OlegGordiichuk:
Add new buttons for token view controller.

Add actions and delegate methods.

### DIFF
--- a/Trust/Tokens/Views/TokensHeaderView.swift
+++ b/Trust/Tokens/Views/TokensHeaderView.swift
@@ -34,10 +34,10 @@ class TokensHeaderView: UIView {
         addSubview(stackView)
 
         NSLayoutConstraint.activate([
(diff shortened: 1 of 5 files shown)


Trust/Tokens/Views/TokensHeaderView.swift
```diff
@@ -9,19 +9,49 @@ class TokensHeaderView: UIView {
         let label = UILabel(frame: .zero)
         label.textColor = Colors.black
         label.textAlignment = .center
+        label.translatesAutoresizingMaskIntoConstraints = false
         return label
     }()
 
+    lazy var sendButton: Button = {
```
vikmeup: Is it possible to reuse view from transactions view controller? Seems like identical to me?

vikmeup:
I'm still considering this, not sure yet what'st the best way to present receive address on token view

PR #427, opened 2018-03-06
Implement HD Wallet

vikmeup:


vikmeup:
@aleph7 can you take a look? I feel like there are some issues with signing transaction. I have ETH on my Ropsten wallet, but it gives an error on Insufficient funds

Imported wallet from Meta Mask! Worked!

vikmeup:
@aleph7 I'm merging this, but mnemonic phrase is disabled for the user, we need to figure those issues before moving forward with bip39.

PR #434, opened 2018-03-08
Add BrowserErrorView

### LINKED ISSUE #363: Present error messages in the webview
We need to have the same functionality as the browser have. 

If you open Trust Browser with internet off it will show alert view, we should show this error inside webview.

hewigovens:
Fix #363, It's actually a overlay view


vikmeup:
@hewigovens if we are adding `ErrorView`, can we have reload/refresh button then? 

can you share a screenshot/gif? curious to see how that looks. I would love to have same functionality as Safari has tho.

hewigovens:
@vikmeup Yes, we can add buttons, I will send you the demo via telegram

vikmeup:
@hewigovens let's add reload button `Button(style: clear)` under error message and center it vertically and should be good to go!

vikmeup:
Looking good @hewigovens !

PR #435, opened 2018-03-08
Update to fix HD wallets

aleph7:


vikmeup:
Nice work @aleph7! Thanks!

All Meta Mask and Ledger users would be very happy!

PR #436, opened 2018-03-08
Add fix for tokens view reload.

OlegGordiichuk:
Add fix for issue with collectable refresh for empty refresh.


Trust/Tokens/ViewControllers/NonFungibleTokensViewController.swift
```diff
@@ -96,7 +96,11 @@ class NonFungibleTokensViewController: UIViewController {
 
     func fetch() {
         startLoading()
-        viewModel.fetchAssets()
+        viewModel.fetchAssets { state in
+            if state {
+                self.endLoading()
```
vikmeup: Why we are not stopping loading if state false?

PR #437, opened 2018-03-08
Fix of the data source.

### LINKED ISSUE #432: Wallets views - offset
If you have two sections in the table view, you say 0 rows for the first section, it add extra spacing, this should be fixed

Current:
simulator screen shot - iphone x - 2018-03-07 at 13 35 15

OlegGordiichuk:
I update data source to respect section count. 

Fix #432

### DIFF
--- a/Trust/Accounts/ViewModels/AccountsViewModel.swift
+++ b/Trust/Accounts/ViewModels/AccountsViewModel.swift
@@ -5,27 +5,30 @@ import TrustKeystore
 
 struct AccountsViewModel {
 
-    private struct Sections {
-        static let hdWallets = 0
-        static let regularWallets = 1
-    }
-
-    private let regularWallets: [Wallet]
-    private let hdWallets: [Wallet]
+    private var walletsTypes = [[Wallet]]()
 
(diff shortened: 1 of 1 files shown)


Trust/Accounts/ViewModels/AccountsViewModel.swift
```diff
@@ -5,70 +5,58 @@ import TrustKeystore
...
-        switch section {
-        case Sections.hdWallets:
-            return hdWallets.isEmpty ? .none : NSLocalizedString("wallet.section.hdWallet.title", value: "HD Wallet", comment: "")
-        case Sections.regularWallets:
-            return .none
-        default: return .none
+        guard walletsTypes.count == 2 && section == 0 else {
+            return nil
         }
+        return NSLocalizedString("wallet.section.hdWallet.title", value: "HD Wallet", comment: "")
```
vikmeup: I think we want to show `HD wallet` if `walletsTypes.count == 1` a well

OlegGordiichuk: Hmm yeah it is true we should have some another way for it

PR #442, opened 2018-03-10
[WIP] Initial implementation order by fiat value (Do not merge)

vikmeup:


vikmeup:
@OlegGordiichuk can you take a look?

OlegGordiichuk:
Hmm I will do this after transaction screen refactoring.

Trust/Tokens/ViewModels/TokensViewModel.swift
```diff
@@ -15,7 +15,7 @@ struct TokensViewModel {
 
     var tokensNetwork: TokensNetworkProtocol
 
-    let tokens: Results<TokenObject>
+    var tokens: [TokenObject] = []
```
OlegGordiichuk: It is a bit wrong way. Hmm we should find more sophisticated way.

vikmeup:
we can leave this for later

PR #445, opened 2018-03-10
Use `statvoo` API to fetch favicons.

VKoskiv:


Trust/Browser/ViewModel/BookmarkViewModel.swift
```diff
@@ -16,6 +16,7 @@ struct BookmarkViewModel {
         return bookmark.title
     }
     var imageURL: URL? {
-        return URL(string: "\(bookmark.url)favicon.ico")
+        let host = URL(string: bookmark.url)?.host
+        return URL(string: "https://api.statvoo.com/favicon/?url=\(host ?? "www.trustwalletapp.com")")
```
OlegGordiichuk: Could we use guard to prevent usage of the ??

Trust/Browser/ViewModel/BookmarkViewModel.swift
```diff
@@ -16,6 +16,7 @@ struct BookmarkViewModel {
         return bookmark.title
     }
     var imageURL: URL? {
-        return URL(string: "\(bookmark.url)favicon.ico")
+        let host = URL(string: bookmark.url)?.host
```
vikmeup: Might be a good idea to add `linkURL: URL?` to the `Bookmark` model, I beelive it is used in another place as well?

Trust/Browser/ViewModel/BookmarkViewModel.swift
vikmeup: 👍 guard host!

Trust/Browser/ViewModel/BookmarkViewModel.swift
VKoskiv: Or change the existing `url` to `URL`?

PR #452, opened 2018-03-11
Issues/559

OlegGordiichuk:
I update view of the network status.

simulator screen shot - iphone x - 2018-03-11 at 18 51 43


Trust/EtherClient/ChainState.swift
```diff
@@ -16,7 +16,7 @@ class ChainState {
         return "\(config.chainID)-" + Keys.latestBlock
     }
 
-    var chainStateCompletion: ((Bool) -> Void)?
+    var chainStateCompletion: ((Bool, Int) -> Void)?
```
vikmeup: I would switch to use `BigInt()` for the future.

PR #453, opened 2018-03-12
Validate mnemonic phrases

### LINKED ISSUE #28: Send ERC20

aleph7:
Fix #28


vikmeup:
@aleph7 oh, that's nice! Do you think it's worth having this check on import functionality as well in the keystore lib itself?

PR #455, opened 2018-03-12
Fix bookmark deletion crash

VKoskiv:


Trust/Browser/ViewControllers/BookmarkViewController.swift
```diff
@@ -88,12 +89,12 @@ extension BookmarkViewController: StatefulViewController {
 
 extension BookmarkViewController: UITableViewDataSource {
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        return viewModel.bookmarks.count
+        return viewModel.bookmarkCount
```
vikmeup: as an alternative call it numberOfRows

PR #475, opened 2018-03-14
Issues/560

vikmeup:


Trust/Tokens/Types/TokensTickerOperation.swift
```diff
@@ -0,0 +1,45 @@
+// Copyright SIX DAY LLC. All rights reserved.
+
+import UIKit
+import TrustKeystore
+
+class TokensTickerOperation: TrustOperation {
+
+    private var network: TokensNetworkProtocol
+
```
vikmeup: can you remove spacing? to keep it smaller

OlegGordiichuk: done

PR #486, opened 2018-03-15
Bump realm schema version

MillerApps:
I bump the schema version for Realm as well as fix the issue of not being able to dismiss the export private key view.

### DIFF
--- a/Trust/Export/Coordinators/ExportPrivateKeyCoordinator.swift
+++ b/Trust/Export/Coordinators/ExportPrivateKeyCoordinator.swift
@@ -35,7 +35,6 @@ class ExportPrivateKeyCoordinator: Coordinator {
     }
 
     func start() {
-        //navigationController.viewControllers = [exportViewController]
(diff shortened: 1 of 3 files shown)


Trust/Export/Coordinators/ExportPrivateKeyCoordinator.swift
```diff
@@ -35,7 +35,7 @@ class ExportPrivateKeyCoordinator: Coordinator {
     }
 
     func start() {
-        navigationController.viewControllers = [exportViewController]
+        navigationController.pushViewController(exportViewController, animated: false)
```
vikmeup: export should done in modal view

vikmeup:
Fixed in here https://github.com/TrustWallet/trust-wallet-ios/commit/e29f6e870ab36aa66eeb5396a1869fde49862d42

MillerApps:
@vikmeup Your solution was much better. At least I can learn from it now.

PR #489, opened 2018-03-15
Attempt to fix issue #485

### LINKED ISSUE #485: Import wallet QR Code Reader not working
The QR Code reader doest work on the import screen. 

To reproduce: Try to import a private key from a QR code.

Note: in my test the stock camera app could successfully read the code.

MillerApps:
An attempt to fix #485


Trust/Wallet/ViewControllers/ImportWalletViewController.swift
```diff
@@ -253,6 +253,7 @@ class ImportWalletViewController: FormViewController {
 
     func setValueForCurrentField(string: String) {
         let type = ImportSelectionType(title: segmentRow?.value)
+        guard let result = QRURLParser.from(string: string) else { return }
```
vikmeup: can you call this function in .address case? 

Nice catch

PR #497, opened 2018-03-18
Issues/483

### LINKED ISSUE #483: We should refactor Transaction and LocalizedOperationObject objects to be able to support Decodable protocol.
Models Transaction and LocalizedOperationObject should support Decodable protocol.

And we should get read of the unnecessary wrappers : RawTransaction, RawLocalizedOperation.

For more info take a look at these classes: TokenObject, NonFungibleTokenObject.

DarthMike:
Hi, here's an attempt to fix #483, refactoring as requested.

I've tried to keep the code style and the same parsing logic. 

There's two things of note:

1. Improved parsing of `to` address in `Transaction` objects: The original parsing in the `from` method of `RawTransaction` struct was only verifying that the `from` address was convertible to `Address` struct, but didn't check for the `to` address. I think it should be verifying both.

2. `ERC20Contract` remains as a temporary struct to parse contract information into the `LocalizedOperationObject`, but it's not used anywhere else.


OlegGordiichuk:
@DarthMike nice work! Thanks for contributing. We also need @vikmeup to review changes.

vikmeup:
Nice work! One thing about to field, it is optional field. When contract created transaction has contract field, instead of to. 

I will be able to give proper code tonight.

DarthMike:
@vikmeup I created a test transaction under Ropsten network to a test contract token, 'to' was getting parsed correctly. Not sure if that is what you mean:
screen shot 2018-03-19 at 16 07 04

vikmeup:
@DarthMike per the documentation on https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sendtransaction
`to: DATA, 20 Bytes - (optional when creating new contract) The address the transaction is directed to.`
We might be setting `to` on `https://github.com/TrustWallet/trust-ray` to be compatible, but this will change.

DarthMike:
I see, I will change shortly for it to be parsed as it was before. Thanks for explanation.

vikmeup:
@DarthMike 🙏! Congrats on your first contribution! Let's merge it once that change is made!

DarthMike:
Great! Commit amended.

PR #498, opened 2018-03-18
Fix memory leaking of the network state view.

OlegGordiichuk:
Fix of the memory leaking of the bar button item.


vikmeup:
@OlegGordiichuk I wonder why `self.session.chainState.chainStateCompletion = { [weak self]` doesn't break retain cycle? 
Is there other solutions to this problem?

Issue #501, opened 2018-03-18
Do not override disabled state for tokens


vikmeup:
@OlegGordiichuk https://github.com/TrustWallet/trust-wallet-ios/blob/ca5615cfe8d7472769dcc4404b37d936709ed642/Trust/Tokens/Types/TokensOperation.swift#L34

If you had a token in the disabled state it will override it and set it as enabled. In the previous version, I had logic that would only update name, symbol, decimals and that's it!

PR #502, opened 2018-03-19
Do not override disabled state for tokens info

vikmeup:
Fixes: https://github.com/TrustWallet/trust-wallet-ios/issues/501


vikmeup:
@OlegGordiichuk found an issue! If you enable/disable tokens they will come back up, because token list all would override old values. 

Added `updateInfo` action to just override specific values that are needed to be updated. 

This should resolve duplicate tokens and enable/disable bug

Issue #477, opened 2018-03-14
Improve progress bar visibility


vikmeup:
The progress bar on browser have the same color as a navigation bar, makes it hard to see.

MillerApps:
I may be mistaken but I don't think you can change the progress bar color of an SFSfariViewController. http://www.openradar.me/22497026

vikmeup:
This is for DApp Browser! we have progress bar in it.

MillerApps:
Oh, yeah that doesn't use the SFSfariViewController.

PR #504, opened 2018-03-19
Change tint of progress bar

MillerApps:
Made the progress bar in the DApp Browser more visible as mentioned in #477.

Color: `#01CDF7`

### DIFF
--- a/Trust/Browser/ViewControllers/BrowserViewController.swift
+++ b/Trust/Browser/ViewControllers/BrowserViewController.swift
@@ -62,7 +62,7 @@ class BrowserViewController: UIViewController {
         let progressView = UIProgressView(progressViewStyle: .default)
         progressView.translatesAutoresizingMaskIntoConstraints = false
         progressView.tintColor = Colors.darkBlue
-        progressView.trackTintColor = .white
+        progressView.trackTintColor = .clear
         return progressView
     }()
 
--- a/Trust/Browser/ViewControllers/BrowserViewController.swift
+++ b/Trust/Browser/ViewControllers/BrowserViewController.swift
@@ -62,6 +62,7 @@ class BrowserViewController: UIViewController {
         let progressView = UIProgressView(progressViewStyle: .default)
         progressView.translatesAutoresizingMaskIntoConstraints = false
         progressView.tintColor = Colors.darkBlue
+        progressView.trackTintColor = .white
         return progressView
     }()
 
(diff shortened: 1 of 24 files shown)


vikmeup:
@MillerApps I don't think it's a solution to the problem.

MillerApps:
@vikmeup What did you have in mind?

vikmeup:
How about making navigation bar white on dapps browser? and keep same progress bar? Similar to Safari?

MillerApps:
@vikmeup Sounds good to me, I'll get it changed tonight.

Trust/Browser/ViewControllers/BrowserViewController.swift
```diff
@@ -114,10 +114,17 @@ class BrowserViewController: UIViewController {
...
 
+        UIApplication.shared.statusBarStyle = .default
         browserNavBar?.browserDelegate = self
         reloadButtons()
     }
 
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+
+        UIApplication.shared.statusBarStyle = .lightContent
```
vikmeup: I don't think it's a good idea to use global status bar setting

Have you tried looking into this: https://stackoverflow.com/a/19014724?

vikmeup:
@MillerApps can you attach a screenshot or gif as well to see how it works?

Trust/Browser/ViewControllers/BrowserViewController.swift
MillerApps: Fixed

Trust/Extensions/UINavigationController.swift
```diff
@@ -8,4 +8,19 @@ extension UINavigationController {
         navigationBar.tintAdjustmentMode = .normal
         navigationBar.tintAdjustmentMode = .automatic
     }
+
+    open override var preferredStatusBarStyle: UIStatusBarStyle {
+        var preferredStyle: UIStatusBarStyle
+        if navigationBar.isKind(of: BrowserNavigationBar.self) {
```
vikmeup: Is there any other ways? Hardcoding this way make it hard to manage cose

vikmeup: Code

MillerApps: Not that I am aware of.

hewigovens: usually return topViewController.preferredStatusBarStyle if you set UIViewControllerBasedStatusBarAppearance: true

MillerApps: Wouldn’t the style be .default, unless each viewController provides a preferredStatusBarStyle of .lightContent?

MillerApps: In order for it to work with topViewController.preferredStatusBarStyle a base UIViewController subclass would need to be used in place of UIViewController. As preferredStatusBarStyle can’t be overridden through an extension on UIViewController. So should this be the solution used?

PR #543, opened 2018-03-26
Add Spotify like tabBar Animation

MillerApps:


vikmeup:
@MillerApps why tho? I don't think it's a nessesary feature atm.

MillerApps:
@vikmeup no reason, just thought it would be a fun addition. 

MillerApps:
Fine by me just messing around. Wasn't sure if it was something that you would like. 

PR #567, opened 2018-04-03
Bug/551

OlegGordiichuk:
- Memory leak fix on edit tokens view controller.
- Fix for crash in edit tokens view controller.
- lazy operations.


Trust/Tokens/ViewModels/EditTokenViewModel.swift
```diff
@@ -108,4 +111,15 @@ class EditTokenViewModel {
         let token = self.token(for: indexPath)
         self.storage.update(tokens: [token.token], action: action)
     }
+
```
vikmeup: @hewigovens can you take a look? This is a hack, but we need better solution

hewigovens: It’s a common workaround, the index somehow is invalid

PR #569, opened 2018-04-04
Improve hex string to data conversion performance

### LINKED ISSUE #563: Freezes when message to sign is too long, what is the limit?
Some of the examples here crash/freeze the Trust iOS app and I think this is primarily because of the length

https://danfinlay.github.io/js-eth-personal-sign-examples/

I like the idea of putting a terms of service into the signing prompt, but using the UIAlert on iOS doesn't seem to be best to handle this

alejandro-isaza:
Fix #563

### DIFF
--- a/Trust/Extensions/Data.swift
+++ b/Trust/Extensions/Data.swift
@@ -12,17 +12,27 @@ extension Data {
     }
 
     init(hex: String) {
-        let len = hex.count / 2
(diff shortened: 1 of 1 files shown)


alejandro-isaza:
Replaced with https://github.com/TrustWallet/trust-wallet-ios/pull/576

PR #570, opened 2018-04-04
Fix tinting issues

MillerApps:
This fixes some `UIActivtityViewController` tint issues. **Please improve upon as you feel fit!**

### DIFF
--- a/Trust/Browser/ViewControllers/BrowserViewController.swift
+++ b/Trust/Browser/ViewControllers/BrowserViewController.swift
@@ -250,13 +250,7 @@ class BrowserViewController: UIViewController {
     private func share() {
         guard let url = webView.url else { return }
         guard let navigationController = navigationController else { return }
-        let activityViewController = ActivityViewController(
-            activityItems: [
-                url,
-                ],
-            applicationActivities: nil,
-            navigation: navigationController
-        )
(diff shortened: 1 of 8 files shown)


Trust/Browser/ViewControllers/BrowserViewController.swift
```diff
@@ -247,17 +247,19 @@ class BrowserViewController: UIViewController {
...
-    }
-
     private func share() {
         guard let url = webView.url else { return }
         guard let navigationController = navigationController else { return }
-        let controller = makeShareController(url: url)
-        controller.popoverPresentationController?.sourceView = navigationController.view
-        controller.popoverPresentationController?.sourceRect = navigationController.view.centerRect
-        navigationController.present(controller, animated: true, completion: nil)
+        let activityViewController = ActivityViewController(
```
vikmeup: can we add navigationController into initializer? so we don't forget to call this method

`makeShareController` can you make other places to use this initializer?

MillerApps: You want to use`makeShareController` for each `ActivityViewController`?

vikmeup: Yeah, that makes it more reusable

MillerApps: Done

PR #595, opened 2018-04-11
Move from completionWithItemsHandler to viewDidDisappear

vikmeup:


Trust/UI/ActivityViewController.swift
```diff
@@ -3,25 +3,32 @@
 import UIKit
 
 class ActivityViewController: UIActivityViewController {
+
+    let navigation: UINavigationController
```
vikmeup: this would be a retain cycle? should be optional then?

PR #608, opened 2018-04-13
Fixed sharing

MillerApps:


Trust/Extensions/UIActivityViewController.swift
```diff
@@ -0,0 +1,9 @@
+// Copyright SIX DAY LLC. All rights reserved.
+
+import UIKit
+
+extension UIActivityViewController {
+    static func makeShareController(items: [Any]) -> UIActivityViewController {
```
vikmeup: can we do just `make()` then?

MillerApps: sure

PR #610, opened 2018-04-13
Fixed Unit Tests/ Travis

MillerApps:
I fixed broken Unit test that were caused by the `NavigationController` class and as well the commenting out the old web3 code.


TrustTests/Factories/WalletSession.swift
```diff
@@ -8,30 +8,30 @@ import BigInt
...
+            //web3: web3,
             balanceCoordinator: balance,
             nonceProvider: GetNonceProvider.make()
         )
     }
     static func makeWithEthBalance(
         account: Trust.Wallet = .make(),
         config: Config = .make(),
-        web3: Web3Swift = Web3Swift(),
+        //web3: Web3Swift = Web3Swift(),
```
vikmeup: can you remove web3 ?

Gemfile.lock
```diff
@@ -1,45 +1,46 @@
 GEM
   remote: https://rubygems.org/
   specs:
```
vikmeup: why gems changed?

MillerApps: Fixed Travis builds.

MillerApps: I ran bundle update.

TrustTests/Factories/WalletSession.swift
MillerApps: Done

PR #633, opened 2018-04-14
493

johnnynanjiang:
Changelog:
1. Replace CoinTicker struct with CoinTickerObject Realm object
2. Add 'balance' field to TokenObject, it gets updated whenever TokenObject, CoinTicker, and Balance get updated.

ToDo:
1. Could use TokenObject.balance in UI (TokensViewController, TokensViewModel, TokesLayout), but that would require some architecture change in order to instantly reflect TokenObject.balance, otherwise, there would be some delay in displaying balance value in UI due to TokenObject being Realm live object.


Podfile.lock
```diff
@@ -203,4 +167,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 7f1175e0e6447ae992c9047de2c88e12564ca350
 
-COCOAPODS: 1.5.0
+COCOAPODS: 1.4.0
```
MillerApps: Can you update to Cocoapods 1.5.0?

Trust/Core/Initializers/MirgrationInitializer.swift
```diff
@@ -36,6 +36,9 @@ class MigrationInitializer: Initializer {
                 fallthrough
             case 33...48:
                 migration.deleteData(forType: Transaction.className)
+                migration.enumerateObjects(ofType: TokenObject.className()) { _, newObject in
```
vikmeup: is there a need for this? balance will be 0 by default

Trust/Settings/Types/Config.swift
```diff
@@ -64,4 +64,8 @@ struct Config {
         get { return defaults.bool(forKey: Keys.testNetworkWarningOff) }
         set { defaults.set(newValue, forKey: Keys.testNetworkWarningOff) }
     }
+
+    var tickersKey: String {
```
vikmeup: Why do you need `tickersKey` of each wallet and network will have its own DB? unless you use shared DB

vikmeup: Can we use shared DB for pricing? since it's not changing

Trust/Tokens/Types/CoinTickerObject.swift
```diff
@@ -0,0 +1,79 @@
...
+        tickersKey: String = "") {
+        self.init()
+        self.id = id
+        self.symbol = symbol
+        self.price = price
+        self.percent_change_24h = percent_change_24h
+        self.contract = contract
+        self.image = image
+        self.tickersKey = tickersKey
+        self.key = "\(self.id)_\(symbol)_\(contract)_\(tickersKey)"
```
vikmeup: that's a huge key - can we make it smaller?

Trust/Tokens/Storage/TokensDataStore.swift
```diff
@@ -118,39 +125,65 @@ class TokensDataStore {
                         "name": token.name,
                         "symbol": token.symbol,
                         "decimals": token.decimals,
+                        "balance": self.getBalance(for: token, with: self.tickers()),
```
vikmeup: I don't think it's needed here

Trust/Core/Network/TrustNetwork.swift
```diff
@@ -42,7 +42,18 @@ class TrustNetwork: NetworkProtocol {
...
-                let tickers = try response.map([CoinTicker].self, atKeyPath: "response", using: JSONDecoder())
+                let rawTickers = try response.map([CoinTickerObject].self, atKeyPath: "response", using: JSONDecoder())
+                let tickers = rawTickers.map {rawTicker in
+                    return CoinTickerObject(
+                        id: rawTicker.id,
+                        symbol: rawTicker.symbol,
+                        price: rawTicker.price,
+                        percent_change_24h: rawTicker.percent_change_24h,
+                        contract: rawTicker.contract,
+                        image: rawTicker.image,
```
vikmeup: can we remove the image from here? it's not being used anymore. also symbol and id not needed as well - we use contract to pull data

vikmeup:
@johnnynanjiang nice work! just some comments and suggestions

Trust/Core/Initializers/MirgrationInitializer.swift
johnnynanjiang: The app will crash without the migration when TokenObject table already has data in it.

Thread 1: Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=io.realm Code=10 "Migration is required due to the following errors:

Trust/Settings/Types/Config.swift
johnnynanjiang: As the price varies for different currency, the tickersKey will look like below:
ticker-aud-1
ticker-usd-1

Trust/Tokens/Types/CoinTickerObject.swift
johnnynanjiang: We could have a discussion here, especially if we want to put tickers in the shared db. Since id, symbol, and contract could be identical for different networks, so we do need tickersKey as it contains network id and currency id .

Trust/Tokens/Storage/TokensDataStore.swift
johnnynanjiang: My misunderstanding, code removed.

Trust/Core/Network/TrustNetwork.swift
johnnynanjiang: Could it be looked at along with putting CoinTickerObject in the shared db, probably a separate issue. As there might be some analysis, as well as regression testing and db migration testing required.

Trust/Core/Initializers/MirgrationInitializer.swift
vikmeup: You need to raise migration ID and you should be fine since you are just adding new fields

Trust/Core/Initializers/MirgrationInitializer.swift
```diff
@@ -36,12 +36,20 @@ class MigrationInitializer: Initializer {
...
-                    newObject?["balance"] = TokenObject.DEFAULT_BALANCE
-                }
+                MigrationInitializer.migrateTokenObjectBalanceField(migration)
             default:
                 break
             }
         }
     }
+
+    static func migrateTokenObjectBalanceField(_ migration: Migration) {
```
vikmeup: don't worry about migration - this is extra code - it will calculate balance for the users once he log in

vikmeup: Default value would be 0

johnnynanjiang: Sure, remove the code. Not sure why I got "migration is required" errors during testing.

Trust/Tokens/Storage/TokensDataStore.swift
```diff
@@ -97,11 +101,14 @@ class TokensDataStore {
...
-                ]
-                realm.create(TokenObject.self, value: update, update: true)
+                guard let tokenObject = realm.object(ofType: TokenObject.self, forPrimaryKey: balance.key.description) else {
+                    continue
+                }
+
+                tokenObject.value = balance.value.description
+                tokenObject.balance = self.getBalance(for: tokenObject, with: self.tickers())
+
+                realm.add(tokenObject, update: true)
```
vikmeup: I think the reasons we call `realm.create(TokenObject.self, value: update, update: true)` with only a few parameters in `update` so we don't override any data, can you adjust to the previous solution? otherwise you will be overriding all the data

Trust/Tokens/Storage/TokensDataStore.swift
```diff
@@ -125,32 +132,57 @@ class TokensDataStore {
...
-    func saveTickers(tickers: [CoinTicker]) {
+    func saveTickers(tickers: [CoinTickerObject]) {
         guard !tickers.isEmpty else {
             return
         }
+
+        deleteAllExistingTickers()
+
+        let coinTickerObjects = tickers.map { (ticker) -> CoinTickerObject in
+            let tickersKey = self.config.tickersKey
```
vikmeup: can you reuse this mapping? you have duplicate code above

Trust/Tokens/Storage/TokensDataStore.swift
```diff
@@ -125,32 +132,57 @@ class TokensDataStore {
...
+
+    private var tickerResultsByTickersKey: Results<CoinTickerObject> {
+        return realm.objects(CoinTickerObject.self).filter("tickersKey == %@", self.config.tickersKey)
     }
 
-    func deleteTickers() {
-        config.defaults.removeObject(forKey: tickersKey)
+    func deleteAllExistingTickers() {
+        do {
+            try realm.write {
```
vikmeup: can you do try? 

I don't think we need to print the error

Trust/Tokens/Types/CoinTickerObject.swift
```diff
@@ -0,0 +1,79 @@
+// Copyright SIX DAY LLC. All rights reserved.
+
+import Foundation
+import Realm
+import RealmSwift
+
+class CoinTickerObject: Object, Decodable {
```
vikmeup: can you keep same name `CoinTicker` ?

Podfile.lock
```diff
@@ -1,6 +1,6 @@
 PODS:
-  - Alamofire (4.6.0)
-  - APIKit (3.1.1):
```
vikmeup: can you avoid updating Pods?

Trust/Tokens/Storage/TokensDataStore.swift
johnnynanjiang: Ok I will change it back. I thought since tokenObject has been retrieved so that I can use add to add it back updated.

vikmeup:
@johnnynanjiang Nice work! can you fix merge conflict and let's merge this!

Trust/Tokens/ViewModels/TokensViewModel.swift
```diff
@@ -156,12 +156,8 @@ class TokensViewModel: NSObject {
 
     private func balances(for tokens: [TokenObject]) {
 
-        tokensTickerOperation.completionBlock = { [weak self] in
-            DispatchQueue.main.async {
-                self?.store.saveTickers(tickers: tokensTickerOperation.tickers)
-                self?.delegate?.refresh()
-            }
-        }
```
vikmeup: is this intended?

PR #643, opened 2018-04-18
Feature/555

OlegGordiichuk:
- Fix app leaking.
- Add promise kit.
- Code refactoring.
- Code simplifying.
- Code cleaning.
- Requests optimization.


Trust/Tokens/ViewModels/TokensViewModel.swift
```diff
@@ -127,77 +115,71 @@ class TokensViewModel: NSObject {
...
-            }
-            self.store.update(balances: [TokensDataStore.etherToken().address: balance.value])
-            completion?(true)
+    func updateEthBalance() {
+        firstly {
+            tokensNetwork.ethBalance()
+        }.done { [weak self] balance in
+            self?.store.update(balances: [TokensDataStore.etherToken().address: balance.value])
+        }.catch { error in
+            print(error)
```
vikmeup: Can we use `NSLog` for consistency? 

What do you think about using Crashlytics logs for this? in that case if something crashes we can see all the logs

Trust/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -231,6 +224,8 @@ extension TokensViewController: UITableViewDataSource {
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         let cell = tableView.dequeueReusableCell(withIdentifier: TokenViewCell.identifier, for: indexPath) as! TokenViewCell
         cell.configure(viewModel: viewModel.cellViewModel(for: indexPath))
+        cell.contentView.isExclusiveTouch = true
```
vikmeup: what is this for?

Trust/Tokens/Storage/TokensDataStore.swift
```diff
@@ -164,3 +179,24 @@ class TokensDataStore {
         )
     }
 }
+
+extension Realm {
```
vikmeup: can we move this to an extension? as a separate file

Would this cause any crash since it's a different thread?

Trust/Core/Network/TrustNetwork.swift
```diff
@@ -1,16 +1,23 @@
...
 
+import PromiseKit
 import Moya
 import TrustCore
 import JSONRPCKit
 import APIKit
 import Result
+import enum Result.Result
+
+enum TrustNetworkProtocolError: Error {
```
vikmeup: can you conform to `LocalizedError`?

Trust/Core/Network/TrustNetwork.swift
```diff
@@ -1,16 +1,23 @@
 // Copyright SIX DAY LLC. All rights reserved.
 
+import PromiseKit
 import Moya
 import TrustCore
 import JSONRPCKit
 import APIKit
 import Result
+import enum Result.Result
```
vikmeup: Strange because `import Result` already imported above

OlegGordiichuk: It is connected with same generic name in promis kit.

Trust/Tokens/ViewControllers/TokensViewController.swift
OlegGordiichuk: To be sure that one tap will open one view controller.

Trust/Tokens/Storage/TokensDataStore.swift
OlegGordiichuk: Ok will do

Trust/Core/Network/TrustNetwork.swift
OlegGordiichuk: Yeah

Trust/Tokens/ViewModels/TokensViewModel.swift
OlegGordiichuk: Done

vikmeup:
@OlegGordiichuk can you rebase and should be good to merge if green

Issue #474, opened 2018-03-13
Full support bytes type / big ints in string for 712


vikmeup:
Based on https://github.com/TrustWallet/trust-wallet-ios/pull/450

alejandro-isaza:
Reference docs: http://solidity.readthedocs.io/en/develop/abi-spec.html#basic-design

vikmeup:
@alejandro-isaza @hewigovens I think [web3swift](https://github.com/BANKEX/web3swift) has this implemented already. We are going to use this lib, does it make sense to bring now?

alejandro-isaza:
I'd rather implement it in Swift, but that's just me :)

hewigovens:
What we implemented now is a subset of solidity types, specifically, represented by JSON values. Because we receive data from DApp Browser (web3.js) and we only support bytes by hex strings.

alejandro-isaza:
Now TrustCore has ABI encoding, what else are we missing for this?

vikmeup:
@alejandro-isaza I bet we just need to adapt it to use inside Trust app, can you take a look? 

Btw, that stuff looks great!

hewigovens:
@alejandro-isaza After encoding, ABIValue (.uint, .int, .bool) data length is fixed (32), EIP 712 encoding for this types are compact / variable. It seems that we can't simply replace with TrustCore(ABIValue)?

PR #646, opened 2018-04-19
+ Add BigInt support for EthTypedData

williamchong007:
please suggest if anything needed to be changed, not quite sure


vikmeup:
@williamchong007 is there a way to implement `bytes` as well? https://github.com/TrustWallet/trust-wallet-ios/issues/474

TrustTests/EtherClient/signTypedMessageTests.swift
```diff
@@ -97,6 +97,26 @@ class signTypedMessageTests: XCTestCase {
...
     }
 
+    func testType_int_Value_bigInt() {
+        // from https://like.co/
+        let typedData = EthTypedData(type: "int256", name: "how much?", value: .string(value: "-57896044618658097711785492504343953926634992332820282019728792003956564819968"))
+        let signResult = keystore.signTypedMessage([typedData], for: account)
+        guard case let .success(data) = signResult else {
+            return XCTFail()
+        }
+        XCTAssertEqual(data.hexEncoded, "0xa1cd93b6dd2a18748993c727c8c9be20a784504b7f1fc9eece228bc5004e805d6a0bcabce62994881b562243ace5cc3ce18bd41302c793dce38c301036af01761c")
```
hewigovens: Could you please check the expected result is same as https://github.com/danfinlay/js-eth-personal-sign-examples ?

hewigovens: I verified.

Trust/EtherClient/Types/EthTypedData.swift
```diff
@@ -90,16 +92,20 @@ struct EthTypedData: Decodable {
                 guard size > 0 else { return Data() }
                 if let uint = UInt64(string) {
                     return getTypedData(for: uint, align: size)
-                } else {
-                    //TODO larger than UInt64, we may need BigInt
+                } else if let bigInt = BigUInt(string) {
+                    let encoder = ABIEncoder()
+                    try! encoder.encode(bigInt)
```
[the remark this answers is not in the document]
williamchong007: `?` should be fine too since it should return Data() by default
updated

williamchong007:
@vikmeup would take some time to dig into the `bytes` flow, maybe I will do it in another PR later?

hewigovens:
@vikmeup @williamchong007 for `bytes`, we only support hex strings now (starts with 0x), Should we support plain strings and numbers as well? It's a bit ambiguous for numbers since we lost js type information.

Another PR is fine.

PR #653, opened 2018-04-21
Add transactionsObservation on viewDidLoad

vikmeup:
Once we create new transaction and place in the store we show badge for Transaction tab, but if we don't have this observation then nothing will be shown.

Can you double check that this won't cause any memory leaks?


OlegGordiichuk:
It is ok but we should do some check after we merge #643

PR #659, opened 2018-04-22
Initial firebase analytics

slgraham:
A few things I'm not 100% sure about:

- I put some of the events in coordinators
- I created both success and failure events, but I'm not sure exactly what the failure cases will tell us since I'm not familiar with what might cause them
- I didn't include a parameter for transaction type in completedTransactionFromBrowser since there wasn't a global variable for type there, but it would be helpful to know when the transaction was a signed vs. sent transaction

Let me know what you think.


MillerApps:
@slgraham can you fix the Swift Lint warnings? Otherwise it looks good to me. 

Trust/Transfer/Coordinators/SendTransactionCoordinator.swift
```diff
@@ -40,8 +40,12 @@ class SendTransactionCoordinator {
                 case .success(let count):
                     let transaction = self.appendNonce(to: transaction, currentNonce: count)
                     self.signAndSend(transaction: transaction, completion: completion)
+                    // analytics event for succesfully sent transaction from wallet
+                    Analytics.track(.sentTransactionFromWallet)
                 case .failure(let error):
                     completion(.failure(AnyError(error)))
+                    // analytics event for failed sent transaction from wallet
+                    Analytics.track(.failedTransactionFromWallet)
```
vikmeup: do you think we should capture error message as well? in that case, you could pass: `error: Error` and then then `error.prettyError` will return friendly error message.

slgraham: Good call - will do.

Do you have any ideas related to my third bullet point in the first comment?

vikmeup: @slgraham once we sign transaction we broadcast it to the network, I think we should be fine for now

PR #671, opened 2018-04-24
update network price comparison (code -> contract)

### LINKED ISSUE #654: Remove comparison by code and use contract instead
`Rate` model contains `code` - this should be removed and any comparison for fetching token price should be done via `contract` to avoid showing incorrect value

hewigovens:
Fix #654

### DIFF
--- a/Trust/Settings/Types/RPCServers.swift
+++ b/Trust/Settings/Types/RPCServers.swift
@@ -75,7 +75,7 @@ enum RPCServer {
         }
     }
 
-    var contract: String {
+    var address: String {
         return "0x0000000000000000000000000000000000000000"
     }
 
--- a/Trust/Settings/Types/RPCServers.swift
+++ b/Trust/Settings/Types/RPCServers.swift
@@ -75,6 +75,10 @@ enum RPCServer {
         }
     }
 
+    var contract: String {
+        return "0x0000000000000000000000000000000000000000"
+    }
+
     var decimals: Int {
         return 18
     }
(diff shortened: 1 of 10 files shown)


Trust/Transactions/ViewModels/BalanceViewModel.swift
```diff
@@ -33,7 +33,7 @@ struct BalanceViewModel: BalanceBaseViewModel {
     var currencyAmount: String? {
         guard let rate = rate else { return nil }
         guard
-            let currentRate = (rate.rates.filter { $0.code == config.server.symbol }.first),
+            let currentRate = (rate.rates.filter { $0.contract == config.server.contract }.first),
```
OlegGordiichuk: This is a weak place contract could be in the lowercase format  but config.server.contract in the uppercase.

vikmeup: I think for the network it will be all the same: `0x0000000000000000000000000000000000000000`

Trust/Settings/Types/RPCServers.swift
```diff
@@ -75,6 +75,10 @@ enum RPCServer {
         }
     }
 
+    var contract: String {
+        return "0x0000000000000000000000000000000000000000"
```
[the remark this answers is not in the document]
hewigovens: I think we will also replace eth contract

TrustTests/Factories/FakeTokensNetwork.swift
```diff
@@ -25,7 +25,7 @@ class FakeTokensNetwork: NetworkProtocol {
     }
     
     func tickers(for tokens: [TokenObject], completion: @escaping ([CoinTicker]?) -> Void) {
-        let eth_contract = "0x0000000000000000000000000000000000000000"
+        let eth_contract = config.server.address
```
vikmeup: can you use `config.server.address` directly? I don't think there is a new to have a separate variable, save below

vikmeup:
Nice work!

PR #672, opened 2018-04-24
Fix 474, support bytes in both hex and plain strings

hewigovens:


Trust/EtherClient/Types/EthTypedData.swift
```diff
@@ -86,7 +86,9 @@ struct EthTypedData: Decodable {
             return getTypedData(for: int, align: size)
         case .string(let string):
             if type.starts(with: "bytes") {
-                return Data(hex: string)
+                if string.isHexEncoded {
+                    return Data(hex: string)
+                }
```
vikmeup: what happens if it's `bytes` and not `isHexEncoded `?

Trust/Extensions/String.swift
```diff
@@ -13,6 +13,17 @@ extension String {
         return data.hexEncoded
     }
 
+    var isHexEncoded: Bool {
```
vikmeup: Wonder if there is better way to handle this

Trust/EtherClient/Types/EthTypedData.swift
hewigovens: fallback to plain string, we have some tests covered.

Trust/Extensions/String.swift
hewigovens: something similar to: https://github.com/ethjs/ethjs-util/blob/a76d4798e18ab78aabd1bb91cef2ebe0b9f2863c/src/index.js#L192

Issue #404, opened 2018-02-25
Add QR Image when sharing your own address


michaelScoff:
From Receive View

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

davecraige:
Good idea

PR #686, opened 2018-04-28
Share QR code when sharing address

MillerApps:
Solves issue #404

### DIFF
--- a/Trust/Transfer/Coordinators/RequestCoordinator.swift
+++ b/Trust/Transfer/Coordinators/RequestCoordinator.swift
@@ -44,7 +44,8 @@ class RequestCoordinator: Coordinator {
     }
 
     @objc func share(_ sender: UIBarButtonItem) {
-        let activityViewController = UIActivityViewController.make(items: [viewModel.shareMyAddressText, requestViewController.imageView.image as Any].compactMap { $0 })
+        let items = [viewModel.shareMyAddressText, requestViewController.imageView.image as Any].compactMap { $0 }
+        let activityViewController = UIActivityViewController.make(items: items)
         activityViewController.popoverPresentationController?.barButtonItem = sender
         navigationController.present(activityViewController, animated: true, completion: nil)
     }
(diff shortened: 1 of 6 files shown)


vikmeup:
What’s the design decision for adding extension to UImage? It should be done via QRGenerator class instead

MillerApps:
Where is the QRGenerator class? Are you saying we should have a new class called QRGenerator ? I added the extension because we generate QRCodes in a few different places, it made sense to cut down on repeating code.

vikmeup:
Using extensions for every single problem is a bad design

MillerApps:
I will continue this on Telegram.

Trust/Transfer/Coordinators/RequestCoordinator.swift
```diff
@@ -44,7 +44,8 @@ class RequestCoordinator: Coordinator {
     }
 
     @objc func share(_ sender: UIBarButtonItem) {
-        let activityViewController = UIActivityViewController.make(items: [viewModel.shareMyAddressText])
+        guard let image = QRGenerator.generate(from: viewModel.myAddressText) else { return }
```
vikmeup: can you take generated image from the view controller? there is no need to generate twice

MillerApps: Sure

vikmeup:
@MillerApps how does it look on the user end when you share it?

MillerApps:
@vikmeup GitHawk Upload by MillerAppsGitHawk Upload by MillerApps

Trust/Transfer/Coordinators/RequestCoordinator.swift
```diff
@@ -44,7 +44,8 @@ class RequestCoordinator: Coordinator {
     }
 
     @objc func share(_ sender: UIBarButtonItem) {
-        let activityViewController = UIActivityViewController.make(items: [viewModel.shareMyAddressText])
+        guard let image = requestViewController.imageView.image else { return }
```
vikmeup: I would not guard this since image potentially could be empty (or slow at generating) for some reasons and then you would block sharing functionality. I would add URL + image and then flat map it.

Issue #311, opened 2018-02-09
Add ENS support


hewigovens:
- [ ] recipient address (.eth)

vikmeup:
@alejandro-isaza if you are making web3swift client can you add a class to resolve ENS name? this would be a very neat feature.

hewigovens:
@vikmeup https://github.com/TrustWallet/trust-core/blob/master/Sources/Solidity/ENSEncoder.swift

vikmeup:
@hewigovens can you pull it? Can we start with doing a local cache of ENS Resolver - then showing naming in the wallets page? 

I think present wallet name would be a good start!

hewigovens:
you mean reverse lookup in wallets page?

vikmeup:
@hewigovens 👍

PR #712, opened 2018-05-09
Add ENS names resolving / lookup

hewigovens:
This is essential for #311


Trust/EtherClient/ENS/ENSClient.swift
```diff
@@ -0,0 +1,66 @@
...
+        )
+        return Promise { seal in
+            Session.send(request) { result in
+                switch result {
+                case .success(let response):
+                    let data = Data(hex: response)
+                    guard data.count == 32 else {
+                        return seal.reject(ENSError.decodeError)
+                    }
+                    let sub = Data(bytes: data.bytes[12...])
```
vikmeup: what does `12...` does?

hewigovens: take last 20 (32 - 12 ) bytes for Address, maybe data.suffix(20) is better?

vikmeup: do you mind adding a comment that this is getting address, otherwise LGTM

hewigovens: done

Trust/Accounts/ViewModels/AccountViewModel.swift
```diff
@@ -11,17 +11,20 @@ struct AccountViewModel {
     let current: Wallet?
     let walletBalance: Balance?
     let server: RPCServer
+    let ensName: String
```
vikmeup: can we pas ENS Model instead?

Trust/Accounts/Views/AccountViewCell.swift
```diff
@@ -12,8 +12,8 @@ class AccountViewCell: UITableViewCell {
     @IBOutlet weak var activeView: UIView!
     @IBOutlet weak var glassesImageView: UIImageView!
     @IBOutlet weak var walletTypeImageView: UIImageView!
-    @IBOutlet weak var addressLable: UILabel!
     @IBOutlet weak var balanceLable: UILabel!
+    @IBOutlet weak var addressLable: UILabel!
```
vikmeup: can you rename `addressLabel` ?

Trust/Accounts/ViewModels/AccountViewModel.swift
hewigovens: we can change this later, a string is enough right now

Trust/Accounts/Views/AccountViewCell.swift
hewigovens: done

PR #714, opened 2018-05-11
Update of the cell ui.

OlegGordiichuk:
- Set cards font sizes from rarebits.
- Set cards content offsets from rarebits


OlegGordiichuk:
simulator screen shot - iphone x - 2018-05-11 at 17 08 45

PR #727, opened 2018-05-20
Refresh ens names for different networks

hewigovens:


Trust/Settings/Types/RPCServers.swift
```diff
@@ -119,6 +119,20 @@ enum RPCServer {
...
+    var ensContract: String? {
+        // https://docs.ens.domains/en/latest/introduction.html#ens-on-ethereum
+        switch self {
+        case .main:
+            return "0x314159265dd8dbb310642f98f50c066173c1259b"
+        case .ropsten:
+            return "0x112234455c3a32fd11230c42e7bccd4a84e02010"
+        case .rinkeby:
+            return "0xe7410170f87102df0055eb195163a03b7f2bff4a"
+        default:
```
vikmeup: make switch exhaustive

Trust/Accounts/ViewModels/AccountsViewController.swift
```diff
@@ -160,6 +160,9 @@ class AccountsViewController: UITableViewController {
     }
 
     private func refreshENSNames() {
+        guard config.server.ensContract != .none else {
+            return
+        }
```
vikmeup: do you think it might be better to handle on ensManager level? it would be just returning default result. because we gonna checking for ENS in different places

Trust/Settings/Types/RPCServers.swift
```diff
@@ -112,13 +112,27 @@ enum RPCServer {
...
             case .sokol: return "https://trust-sokol.herokuapp.com"
-            case .custom(let custom):
+            case .custom:
                 return "" // Enable? make optional
             }
         }()
         return URL(string: urlString)!
     }
 
+    var ensContract: String {
```
vikmeup: Can you make this `Address?`

Trust/EtherClient/ENS/ENSManager.swift
```diff
@@ -9,13 +9,17 @@ struct ENSManager {
...
+    let client: ENSClient
 
-    init(realm: Realm) {
+    init(realm: Realm, config: Config = Config()) {
         self.store = ENSStore(realm: realm)
+        self.client = ENSClient(server: config.server)
     }
 
     func resolve(name: String, ignoreCache: Bool = false) -> Promise<Address> {
+        guard client.ensAvailable == true else {
```
vikmeup: `client.ensAvailable == true` => `client.ensAvailable` ?

Trust/Settings/Types/RPCServers.swift
hewigovens: Also import TrustCore here?

vikmeup: Yeah

PR #728, opened 2018-05-20
Fix issue with CoinTicker price

johnnynanjiang:
- fix an issue where duplicate CoinTicker records (the ones with old key) caused Token balance being incorrect

### DIFF
--- a/Trust/Tokens/Types/CoinTicker.swift
+++ b/Trust/Tokens/Types/CoinTicker.swift
@@ -25,11 +25,7 @@ class CoinTicker: Object, Decodable {
         self.percent_change_24h = percent_change_24h
         self.contract = contract
         self.tickersKey = tickersKey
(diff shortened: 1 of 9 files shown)


TrustTests/Tokens/TokensDataStoreTest.swift
```diff
@@ -4,14 +4,17 @@ import XCTest
 @testable import Trust
 
 class TokensDataStoreTest: XCTestCase {
+    var config: Config!
```
vikmeup: Make each test independent! Don’t add it in here

Trust/Tokens/Types/CoinTicker.swift
```diff
@@ -69,3 +65,13 @@ extension CoinTicker {
         )
     }
 }
+
+struct CoinTickerKeyFactory {
```
vikmeup: Take a look at examples how we create all the factories. Tests/Factories

PR #730, opened 2018-05-20
Bug/tokens separators

OlegGordiichuk:
- Update of the cell style
- Add new view to show separator
- Remove of the calculation for intents (UIKit constraints will do it for as)
- Set proper offset of the separator.

In this fix separators will be same in all iOS and devices versions.

### DIFF
--- a/Trust/Style/Styles.swift
+++ b/Trust/Style/Styles.swift
@@ -39,6 +39,7 @@ struct StyleLayout {
     struct TableView {
         static let heightForHeaderInSection: CGFloat = 30
         static let separatorColor = UIColor(hex: "d7d7d7")
+        static let separatorHeigh: CGFloat = 0.5
     }
 
     struct CollectibleView {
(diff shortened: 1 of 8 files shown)


Trust/Transactions/ViewControllers/TransactionsViewController.swift
```diff
@@ -149,6 +150,11 @@ class TransactionsViewController: UIViewController {
         }, selector: #selector(Operation.main), userInfo: nil, repeats: true)
     }
 
+    private func showSeparator(for index: IndexPath) -> Bool {
```
vikmeup: @OlegGordiichuk do you think we should move this into `TableView`, let's have our own class for that.

Trust/Style/Styles.swift
```diff
@@ -39,6 +39,7 @@ struct StyleLayout {
     struct TableView {
         static let heightForHeaderInSection: CGFloat = 30
         static let separatorColor = UIColor(hex: "d7d7d7")
+        static let separatorHeigh: CGFloat = 0.5
```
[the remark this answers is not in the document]
OlegGordiichuk: done

Trust/Transactions/ViewControllers/TransactionsViewController.swift
OlegGordiichuk: Done

vikmeup:
@OlegGordiichuk is there any ways to make it work for iOS 11/12 by using native functionality? I'm not fan of custom solutions

PR #734, opened 2018-05-22
rename tickersKey to currencyKey in CoinTicker

johnnynanjiang:


Trust/Core/Initializers/MirgrationInitializer.swift
```diff
@@ -36,9 +36,18 @@ class MigrationInitializer: Initializer {
...
             default:
                 break
             }
         }
     }
 }
+
+class RealmMigration {
+    class func renameTickerCurrencyKey(_ migration: Migration) {
+        migration.renameProperty(onType: "CoinTicker", from: "tickersKey", to: "currencyKey")
```
vikmeup: can you just delete `CoinTicker ` model for simplicity?

Trust/Core/Initializers/MirgrationInitializer.swift
```diff
@@ -36,9 +36,18 @@ class MigrationInitializer: Initializer {
                 fallthrough
             case 33...49:
                 migration.deleteData(forType: Transaction.className)
+                fallthrough
+            case 50...52:
+                RealmMigration.renameTickerCurrencyKey(migration)
```
vikmeup: you need to `fallthrough`

vikmeup:
I don't think we need to rename at all.

johnnynanjiang:
I thought it would be ideal to rename tickersKey to currencyKey to avoid confusion, but it is ok to leave it as is.

PR #739, opened 2018-05-30
Clean up CoinTicker data

johnnynanjiang:


Trust/Core/Initializers/MirgrationInitializer.swift
```diff
@@ -20,7 +20,7 @@ class MigrationInitializer: Initializer {
     }
 
     func perform() {
-        config.schemaVersion = 52
```
vikmeup: can you make one variable for both `schemaVersion`?

Trust/Core/Initializers/MirgrationInitializer.swift
```diff
@@ -36,6 +36,8 @@ class MigrationInitializer: Initializer {
                 fallthrough
             case 33...49:
                 migration.deleteData(forType: Transaction.className)
+            case 50...52:
```
vikmeup: can you use `fallthrough` and 0...52`? just in case

PR #744, opened 2018-06-04
Bug/ui fixes

OlegGordiichuk:
- Remove of additional constraints.
- Clean code.
- Remove warnings.

### DIFF
--- a/Trust/Accounts/Views/AccountViewCell.swift
+++ b/Trust/Accounts/Views/AccountViewCell.swift
@@ -8,7 +8,7 @@ protocol AccountViewCellDelegate: class {
 }
 
 class AccountViewCell: UITableViewCell {
-    
+
     @IBOutlet weak var infoButton: UIButton!
     @IBOutlet weak var activeView: UIView!
     @IBOutlet weak var glassesImageView: UIImageView!
--- a/Trust/Accounts/Views/AccountViewCell.swift
+++ b/Trust/Accounts/Views/AccountViewCell.swift
@@ -8,6 +8,7 @@ protocol AccountViewCellDelegate: class {
 }
 
 class AccountViewCell: UITableViewCell {
+    
     @IBOutlet weak var infoButton: UIButton!
     @IBOutlet weak var activeView: UIView!
     @IBOutlet weak var glassesImageView: UIImageView!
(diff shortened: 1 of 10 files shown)


Trust/Accounts/Views/AccountViewCell.swift
```diff
@@ -8,14 +8,17 @@ protocol AccountViewCellDelegate: class {
 }
 
 class AccountViewCell: UITableViewCell {
+
```
vikmeup: is this spacing intended?

Trust/Tokens/ViewControllers/EditTokensViewController.swift
```diff
@@ -21,12 +21,12 @@ class EditTokensViewController: UITableViewController {
         let searchController = UISearchController(searchResultsController: self.searchResultsController)
         searchController.delegate = self
         searchController.searchResultsUpdater = self
+        searchController.hidesNavigationBarDuringPresentation = true
```
vikmeup: would this break anything else?

Trust/Tokens/ViewModels/NonFungibleTokenViewModel.swift
```diff
@@ -37,6 +37,10 @@ class NonFungibleTokenViewModel {
         return !tokens.isEmpty
     }
 
+    var cellHeight: CGFloat {
+        return 263
```
vikmeup: Is it intended change? it was 250 points

Trust/Accounts/Views/AccountViewCell.swift
OlegGordiichuk: Unfortunately in this class we have issue with spacings. So yes.

Trust/Tokens/ViewControllers/EditTokensViewController.swift
OlegGordiichuk: Nop, it is safe.

Trust/Tokens/ViewModels/NonFungibleTokenViewModel.swift
OlegGordiichuk: Yeah but we add custom offsets and bounds. And in this case to give an option to build proper cell size and silent warning we should set this height.

vikmeup: Can you reduce to 250? or even 240, 263 seems too big

OlegGordiichuk: ok

PR #746, opened 2018-06-06
Set priority for the spacerWidth view.

OlegGordiichuk:
Fix for the spacerWidth method to return view with suitable priority to suspend warnings.

### DIFF
--- a/Trust/Extensions/UIView.swift
+++ b/Trust/Extensions/UIView.swift
@@ -40,7 +40,7 @@ extension UIView {
         return view
     }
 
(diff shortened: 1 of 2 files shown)


Trust/Extensions/UIView.swift
```diff
@@ -40,15 +40,15 @@ extension UIView {
         return view
     }
 
-    static func spacerWidth(_ width: CGFloat = 1, backgroundColor: UIColor = .clear, alpha: CGFloat = 1) -> UIView {
```
vikmeup: maybe you can pass `UILayoutPriority` value instead?

PR #759, opened 2018-06-12
Bug/transactions screen ui fixes

BloodyPixy:
Fixed Loading and Empty views for Transactions screen by changing inset value. Added Retry button for Empty case on transactions screen to gain a refresh ability.

### DIFF
--- a/Trust/Tokens/ViewControllers/TokenViewController.swift
+++ b/Trust/Tokens/ViewControllers/TokenViewController.swift
@@ -21,7 +21,7 @@ class TokenViewController: UIViewController {
     }()
 
     private var insets: UIEdgeInsets {
-        return UIEdgeInsets(top: header.frame.height + 100, left: 0, bottom: 60, right: 0)
+        return UIEdgeInsets(top: header.frame.height + 100, left: 0, bottom: 0, right: 0)
     }
 
     private var viewModel: TokenViewModel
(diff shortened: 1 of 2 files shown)


BloodyPixy:
@vikmeup checked on iOS 10.3.1 (iPhone 6), looks good. The purpose of previous numbers is unclear as there are magic numbers. `ButtonSize.extraLarge.height` - it is unclear which button it is exactly, because on the Empty case view there was none before I have added the closure for `onRetry` case. I think this inset was created to separate the Error, Loading and Empty views from other elements that exist on this screen. Attaching the video samples (totally looks better now and works). There still might be an issue on smaller iPhone 5s screens (and it was there before the fix). I can check it when the simulators will finish downloading. 

[video_samples.zip](https://github.com/TrustWallet/trust-wallet-ios/files/2093567/video_samples.zip)

OlegGordiichuk:
@BloodyPixy thx for the contribution. Could you pls also screens that have same logic for displaying empty table view. 

They are:

- Token Details
- Transactions
- Collectibles

BloodyPixy:
@OlegGordiichuk , I have done what you have asked for, although there is a problem with landscape mode on all the screens mentioned. As the  Error, Loading and Empty views are not in a Table View hierarchy as a cells, I believe, there is a need to recalculate whole layout when changing device orientation (it always was this way) and also consider small devices (like 5s). For now it looks good and works in portrait.

OlegGordiichuk:
@vikmeup this fix will fix ui problems of empty screen. Seam to me code is legimate.

Issue #763, opened 2018-06-13
New endpoint for unregister device available


kolya182:
Please see list of small changes here https://github.com/TrustWallet/trust-ray/pull/326/

PR #775, opened 2018-06-26
Move to POST instead of DELETE on notifications unregister API call

BloodyPixy:
Referencing this https://github.com/TrustWallet/trust-wallet-ios/issues/763 ticket.

### DIFF
--- a/Trust/EtherClient/TrustClient/TrustClient.swift
+++ b/Trust/EtherClient/TrustClient/TrustClient.swift
@@ -45,7 +45,7 @@ extension TrustService: TargetType {
         case .getTokens: return .get
         case .getTransaction: return .get
         case .register: return .post
-        case .unregister: return .delete
+        case .unregister: return .post
         case .marketplace: return .get
         case .assets: return .get
         case .search: return .get
--- a/Trust/Settings/Coordinators/PushNotificationsRegistrar.swift
+++ b/Trust/Settings/Coordinators/PushNotificationsRegistrar.swift
@@ -33,19 +33,7 @@ class PushNotificationsRegistrar {
             preferences: NotificationsViewController.getPreferences()
         )
 
-        trustProvider.request(.unregister(device: device)) { response in
-            guard response.error == nil else {
-                return
-            }
-            if let statusCode = response.value?.statusCode {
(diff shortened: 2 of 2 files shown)


Trust/Settings/Coordinators/PushNotificationsRegistrar.swift
```diff
@@ -33,7 +33,20 @@ class PushNotificationsRegistrar {
             preferences: NotificationsViewController.getPreferences()
         )
 
-        trustProvider.request(.unregister(device: device)) { _ in }
+        trustProvider.request(.unregister(device: device)) { response in
+            guard response.error == nil else {
```
vikmeup: just call `trustProvider.request(.unregister(device: device))`, no need to handle response

Trust/EtherClient/TrustClient/TrustClient.swift
```diff
@@ -45,7 +45,7 @@ extension TrustService: TargetType {
         case .getTokens: return .get
         case .getTransaction: return .get
         case .register: return .post
-        case .unregister: return .delete
+        case .unregister: return .post
```
vikmeup: @kolya182 what's the reason to use `post` instead of `delete`?

kolya182: @vikmeup Following REST API rules for`delete` end point should looks like `DELETE` push/unregister/{TOKEN} in this case we would expose TOKEN in logs, which is not ideal. So as a trade off we choose `post` which is not ideal either

Trust/Settings/Coordinators/PushNotificationsRegistrar.swift
kolya182: @vikmeup What if server is down or user has bad internet connection ?. Hi would think subscription went fine, but in really not.

vikmeup: it's ok. no need to handle it. Let's assume it always succeed

BloodyPixy:
@vikmeup is this PR going to be approved or closed?

vikmeup:
@BloodyPixy nice work! closed

Issue #782, opened 2018-06-28
Show QR Code scanner as push view, not modal


vikmeup:
Push makes more sense for UX

This might need to require to add support for pushing coordinators

vikmeup:
Also, make sense to use `ScanQRCodeCoordinator` for all the QR Code scanning to be consistent

PR #789, opened 2018-06-29
Show QR Code scanner as push view, not modal (#782)

BloodyPixy:
@vikmeup , referencing https://github.com/TrustWallet/trust-wallet-ios/issues/782 issue. I have started to alter the code. I have moved logic for creating and navigating QR Code scanner from `SendViewController` to `SendCoordinator`, since it seems to me, it is coordinators responsibility, and then having a controller reference, I update it using a new dedicated function.

I will proceed to implement this change in all the other places. If you have any comments, feel free.

Edit: I could not test a browser, because it is not working right now.


OlegGordiichuk:
@BloodyPixy could you pls fix lint warnings.

BloodyPixy:
@OlegGordiichuk , I am not done yet.

BloodyPixy:
@OlegGordiichuk , @vikmeup , I have more or less finalized this. In some cases, this is impossible to handle via coordinator, as there in no any.

vikmeup:
@BloodyPixy You gotta build a generic solution to push coordinator, I can work on this later, but there is an examples how it's done

vikmeup:
@BloodyPixy take a look at this: https://irace.me/navigation-coordinators, we only need logic for `pushCoordinator()` make it as protocol?

BloodyPixy:
@vikmeup , check new commits. I believe that's much better and can be applied to all coordinators.

Trust/UI/NavigationController.swift
```diff
@@ -35,3 +37,26 @@ class NavigationController: UINavigationController {
...
+        viewControllersToChildCoordinators[coordinator.rootViewController] = coordinator
+        pushViewController(coordinator.rootViewController, animated: true)
+    }
+    func popCoordinator(_ coordinator: RootCoordinator) {
+        viewControllersToChildCoordinators.removeValue(forKey: coordinator.rootViewController)
+        popViewController(animated: true)
+    }
+}
+
+protocol RootViewControllerProvider: class {
```
vikmeup: can you move into separate file?

vikmeup:
@BloodyPixy Nice work! I will take another look tomorrow!

Trust/Tokens/ViewControllers/NewTokenViewController.swift
```diff
@@ -166,16 +168,14 @@ class NewTokenViewController: FormViewController {
...
 }
 
-extension NewTokenViewController: QRCodeReaderDelegate {
-    func readerDidCancel(_ reader: QRCodeReaderViewController!) {
-        reader.stopScanning()
-        reader.dismiss(animated: true, completion: nil)
+extension NewTokenViewController: ScanQRCodeCoordinatorDelegate {
+    func didCancel(in coordinator: ScanQRCodeCoordinator) {
+        navigationController?.setNavigationBarHidden(false, animated: false)
+        (navigationController as? NavigationController)?.popCoordinator(coordinator)
```
vikmeup: navigationController should be `NavigationController` in here?

Trust/UI/NavigationController.swift
BloodyPixy: Even in two

Trust/Tokens/ViewControllers/NewTokenViewController.swift
BloodyPixy: @vikmeup I know what you are talking about, but in our case, we don't always have a coordinator. Take a look on how NewTokenViewController & ImportWalletViewController are created. They don't have coordinators, but somehow, they need to present the QR Scanner. So it seems to me as a viable solution. Please, let me know if I am missing something.

vikmeup: View controllers can call delegate to the coordinator and you push from the coorditor then, that would be the right way to do it!

BloodyPixy: I agree. On it.

vikmeup: Awesome, this is gonna be a big improvement to our codebase once it’s done

Trust/Browser/Coordinators/BrowserCoordinator.swift
```diff
@@ -41,6 +41,9 @@ class BrowserCoordinator: NSObject, Coordinator {
         controller.delegate = self
         return controller
     }()
+    var providedRootController: UIViewController {
```
vikmeup: why `providedRootController`? just call it `rootViewController`?

Trust/Foundation/PushableCoordinator.swift
```diff
@@ -1,10 +1,44 @@
...
+                return
+            }
+            addCoordinator(coordinator)
+            coordinator.providedRootController.hidesBottomBarWhenPushed = tabBarHidden
+            nvc.setNavigationBarHidden(navigationBarHidden, animated: false)
+            nvc.pushViewController(coordinator.providedRootController, animated: true)
+        }
+    }
+    func popCoordinator(_ coordinator: RootCoordinator, navigationBarHidden: Bool = false) {
+        if let presentedNVC = providedRootController.presentedViewController as? UINavigationController {
```
vikmeup: why do you need to pop? it should be done automatically once view is closed

Trust/Browser/Coordinators/BrowserCoordinator.swift
BloodyPixy: +, it intersected with the existing name of controllers in some places, but I will just change them as well.

OlegGordiichuk:
@BloodyPixy could you pls resolve conflicts?

BloodyPixy:
Closing this PR, as there was a lot of discussion and changes involved. Creating new branch for this issue.

Issue #783, opened 2018-06-28
Generate QR code quicker


vikmeup:
The current solution is too slow! It should be instant

BloodyPixy:
@vikmeup in most cases the QR is generated in less than 0.6 sec. Are we talking about iOS 11.4 or 12? Could you please add some more information regarding the issue and the way to reproduce?

vikmeup:
@BloodyPixy 0.6 is a lot! This should be instant. Let's try `.userInitiated`

BloodyPixy:
I will use `.userInteractive` - "returns a global concurrent queue which is suitable for task that need to be done "quickly" in order to update the user interface. Tasks on this queue have a higher priority than e.g. tasks on a  QOS_CLASS_BACKGROUND queue. But this is not the main queue." I believe it is what we need and it seems to work much faster on real iPhone 5s (Not worth comparing on simulator as it uses hardware acceleration of Mac).

PR #793, opened 2018-07-01
Change background queue to userInterective when generating QR.

BloodyPixy:
Referencing  https://github.com/TrustWallet/trust-wallet-ios/issues/783 issue.


BloodyPixy:
@vikmeup Simple answer: YES! My iphone 5s now generated QR almost instantly!

PR #824, opened 2018-07-09
Update of the pod file.

OlegGordiichuk:
I update pod file probably it is because of the RealmSwift. If it is not solution pls comment whole statement in pod file:
/*
    if target.name != 'Realm'
        target.build_configurations.each do |config|
            config.build_settings['MACH_O_TYPE'] = 'staticlib'
        end
    end
*/


vikmeup:
@OlegGordiichuk didn't help. closing this.

PR #827, opened 2018-07-13
Scroll optimization

OlegGordiichuk:
1) Optimization of the hard balance calculations. (Now in background thread)
2) Remove of the old code that was just wasting of the memory.
3) Optimization of the tickers usage. (We should firstly select all tickers for specific wallet and than sort them. In this case we select tickers in O(1) was O(N) complexity) 
4) Reduce usage of the main thread .


Trust/Tokens/Storage/TokensDataStore.swift
```diff
@@ -72,11 +75,18 @@ class TokensDataStore {
         add(tokens: [newToken])
     }
 
+    func preparedTickres() -> [CoinTicker] {
```
vikmeup: `preparedTickres ` => `preparedTickers `

Trust/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -49,13 +48,16 @@ final class TokensViewController: UIViewController {
         return footerView
     }()
 
+    let tableView: UITableView = {
```
vikmeup: if you initially it here, can you move all the configuration into this block?

Trust/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -158,8 +153,9 @@ final class TokensViewController: UIViewController {
     }
 
     @objc func didBecomeActive() {
-        sheduleBalanceUpdate()
         startTokenObservation()
+        fetch()
+        sheduleBalanceUpdate()
```
vikmeup: `sheduleBalanceUpdate ` => `scheduleBalanceUpdate `

Trust/Transfer/ViewModels/SendViewModel.swift
```diff
@@ -154,14 +154,14 @@ struct SendViewModel {
     }
     /// Get pair price with ticker
     func currentPairPrice() -> Decimal? {
-        guard let currentTokenInfo = storage.tickers().first(where: { $0.contract == destinationAddress.description }), let price = Decimal(string: currentTokenInfo.price) else {
+        guard let currentTokenInfo = storage.tickers.first(where: { $0.contract == destinationAddress.description }), let price = Decimal(string: currentTokenInfo.price) else {
```
vikmeup: can you move this into one function? this this logic reused below as well

Trust/Tokens/Storage/TokensDataStore.swift
OlegGordiichuk: done

Trust/Tokens/ViewControllers/TokensViewController.swift
OlegGordiichuk: yeah totally

Trust/Tokens/ViewControllers/TokensViewController.swift
OlegGordiichuk: done

Trust/Transfer/ViewModels/SendViewModel.swift
OlegGordiichuk: done

Trust/Tokens/ViewControllers/TokensViewController.swift
OlegGordiichuk: done

PR #834, opened 2018-07-23
Replace injection of the cell view model in more appropriate place.

OlegGordiichuk:
Based on the article about smooth scrolling we should bind data tableView:willDisplayCell:forRowAtIndexPath but not at cellForRowAtIndexPath.

Original source:
https://medium.com/ios-os-x-development/perfect-smooth-scrolling-in-uitableviews-fd609d5275a5

### DIFF
--- a/Trust/Tokens/ViewControllers/TokensViewController.swift
+++ b/Trust/Tokens/ViewControllers/TokensViewController.swift
@@ -200,13 +200,15 @@ extension TokensViewController: UITableViewDelegate {
 extension TokensViewController: UITableViewDataSource {
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         let cell = tableView.dequeueReusableCell(withIdentifier: TokenViewCell.identifier, for: indexPath) as! TokenViewCell
-        cell.configure(viewModel: viewModel.cellViewModel(for: indexPath))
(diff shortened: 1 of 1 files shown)


Trust/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -200,14 +200,16 @@ extension TokensViewController: UITableViewDelegate {
 extension TokensViewController: UITableViewDataSource {
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         let cell = tableView.dequeueReusableCell(withIdentifier: TokenViewCell.identifier, for: indexPath) as! TokenViewCell
-        cell.configure(viewModel: viewModel.cellViewModel(for: indexPath))
-        cell.contentView.isExclusiveTouch = true
         cell.isExclusiveTouch = true
```
vikmeup: @OlegGordiichuk `isExclusiveTouch ` is this still required?

vikmeup:
@OlegGordiichuk can you take a look at merged conflict?

Trust/Tokens/ViewControllers/TokensViewController.swift
OlegGordiichuk: @vikmeup yeah to avoid multiple redirections from one cell.

PR #836, opened 2018-07-23
Memory fixes

OlegGordiichuk:
Fix usage of the strong self. 
Manually clean coordinators on app restart.


Trust/Settings/Coordinators/SettingsCoordinator.swift
```diff
@@ -120,6 +120,7 @@ extension SettingsCoordinator: WalletsCoordinatorDelegate {
     }
 
     func didSelect(wallet: WalletInfo, in coordinator: WalletsCoordinator) {
+        coordinator.navigationController.removeChildCoordinators()
```
vikmeup: Any chance we can call this on `InCoordinator`?  on every restart?

PR #841, opened 2018-07-24
Update of the total balance in bg thread.

OlegGordiichuk:
Calculation of the total balance will be done in bg thread.

### DIFF
--- a/Trust/Tokens/ViewControllers/TokensViewController.swift
+++ b/Trust/Tokens/ViewControllers/TokensViewController.swift
@@ -18,7 +18,6 @@ final class TokensViewController: UIViewController {
 
     lazy var header: TokensHeaderView = {
         let header = TokensHeaderView(frame: .zero)
-        header.amountLabel.text = viewModel.headerBalance
         header.amountLabel.textColor = viewModel.headerBalanceTextColor
         header.backgroundColor = viewModel.headerBackgroundColor
         header.amountLabel.font = viewModel.headerBalanceFont
@@ -116,7 +115,9 @@ final class TokensViewController: UIViewController {
     }
 
     func refreshHeaderView() {
-        header.amountLabel.text = viewModel.headerBalance
+        viewModel.amount(completion: { [weak self] value in
+           self?.header.amountLabel.text = value
+        })
     }
 
     @objc func missingToken() {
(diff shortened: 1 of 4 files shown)


Trust/Tokens/ViewModels/TokensViewModel.swift
```diff
@@ -85,20 +81,25 @@ final class TokensViewModel: NSObject {
...
-        guard let coinTicker = store.coinTicker(by: token.address) else {
-            return 0
+    func amount(completion: @escaping (String?) -> Void) {
+        DispatchQueue.global(qos: .background).async { [weak self] in
+          guard let strongSelf = self else {
+                completion(CurrencyFormatter.formatter.string(from: NSNumber(value: TokenObject.DEFAULT_BALANCE)))
+                return
+          }
+
+          let realm = try! Realm(configuration: strongSelf.store.realm.configuration)
```
vikmeup: what's the reason for initializing realm? Can you just `use store.realm`?

vikmeup: If it's shared, then can you pass separate instance?

Trust/Tokens/ViewControllers/TokensViewController.swift
```diff
@@ -116,7 +115,9 @@ final class TokensViewController: UIViewController {
     }
 
     func refreshHeaderView() {
-        header.amountLabel.text = viewModel.headerBalance
+        viewModel.amount(completion: { [weak self] value in
```
vikmeup: would this update total amount on Token view screen?

Trust/Tokens/Views/TokensHeaderView.swift
```diff
@@ -9,6 +9,7 @@ final class TokensHeaderView: UIView {
         let label = UILabel(frame: .zero)
         label.textColor = Colors.black
         label.textAlignment = .center
+        label.text = CurrencyFormatter.formatter.string(from: NSNumber(value: TokenObject.DEFAULT_BALANCE))
```
vikmeup: Can this be set from the `viewModel`?

Trust/Tokens/ViewModels/TokensViewModel.swift
OlegGordiichuk: We should create separate Realm instance in specific bg thread.

Trust/Tokens/ViewControllers/TokensViewController.swift
OlegGordiichuk: Yeah it will

Trust/Tokens/Views/TokensHeaderView.swift
OlegGordiichuk: done

Trust/Tokens/ViewModels/TokensViewModel.swift
vikmeup: Can you remove `!` - I hope this won't happen, but ideally, we need to reduce the number of `!` in the code

PR #849, opened 2018-07-25
Refactoring balance coordinator

OlegGordiichuk:
- Remove of the old code.
- Use recently added function to fulfill all required functionality.

@vikmeup I completely remove old logic that was redundant and use new approach that we start using  recently.


Trust/Core/ViewModels/GasViewModel.swift
```diff
@@ -6,18 +6,18 @@ import BigInt
...
     let server: RPCServer
-    let currencyRate: CurrencyRate?
+    let session: WalletSession
     let formatter: EtherNumberFormatter
 
     init(
         fee: BigInt,
         server: RPCServer,
-        currencyRate: CurrencyRate? = nil,
+        session: WalletSession,
```
vikmeup: can you pass `TokenStorage` only?

This would make testing easier

Trust/Core/ViewModels/GasViewModel.swift
```diff
@@ -27,12 +27,15 @@ struct GasViewModel {
...
+        guard let price = session.tokensStorage.coinTicker(by: server.priceID)?.price else {
+            return .none
+        }
+        return FeesCalculations.estimate(fee: formatter.string(from: fee), with: price)
     }
 
     var monetaryFee: String? {
         guard let feeInCurrency = feeCurrency,
-            let fee = currencyRate?.format(fee: feeInCurrency) else {
+            let fee = FeesCalculations.format(fee: feeInCurrency) else {
```
vikmeup: Call it `FeeCalculations`

TrustTests/Core/ViewModels/GasViewModelTest.swift
```diff
@@ -0,0 +1,15 @@
...
+@testable import Trust
+
+class GasViewModelTest: XCTestCase {
+
+    let viewModel = GasViewModel.make()
+
+    func testEtherFee() {
+        let expectedResult = "\(viewModel.formatter.string(from: viewModel.fee)) \(viewModel.server.symbol)"
+        let result = viewModel.etherFee
+        XCTAssertEqual(expectedResult, result)
```
vikmeup: Is there a way to know the result? I'm not really sure how result looks like

PR #859, opened 2018-07-28
Update for new keystore

alejandro-isaza:


vikmeup:
@alejandro-isaza should we bring new changes? Super excited for upcoming Bitcoin integration

PR #866, opened 2018-07-30
Add new line with wallet name on Transaction Confirmation

BloodyPixy:
Referencing #862


OlegGordiichuk:
@BloodyPixy could you pls add image of the new UI?

Trust/Transfer/ViewModels/ConfirmPaymentDetailsViewModel.swift
```diff
@@ -53,6 +56,16 @@ struct ConfirmPaymentDetailsViewModel {
         return transaction.gasLimit
     }
 
+    var walletNameTitle: String {
+        return NSLocalizedString("confirmPayment.walletTitle.label.title", value: "Wallet Title", comment: "")
+    }
+
+    var currentWalletName: String {
+        return keystore.recentlyUsedWallet?.info.name
```
OlegGordiichuk: We could use keystore.mainWallet instead of the keystore.recentlyUsedWallet it will increase readability of the code. Also could we avoid of "??" it will dramatically increase complication time.

BloodyPixy: keystore.mainWallet is also optional, so we won't avoid '??', but I agree regarding clarity and will use it instead.

BloodyPixy:
@OlegGordiichuk 
img_0589

vikmeup:
@BloodyPixy Ideally it should be: Wallet #3 (0x123...123)

Use your judgment

vikmeup:
@vikmeup Wallet Title => Wallet

BloodyPixy:
@vikmeup , I did it this way, because the sender address (current wallet) is included in 'Sender' field. I suggest merging sender and wallet name then. It will look more clear.

vikmeup:
@BloodyPixy For sure! It should be just one field

BloodyPixy:
@vikmeup let do it like this: 

**From Wallet** 
*Wallet Name* (*address*)

vikmeup:
@BloodyPixy 👍 nice!

BloodyPixy:
@OlegGordiichuk keystore.mainWallet.address is wrong. It doesn't provide the current main wallet address. What does it return us?

Trust/Transfer/ViewModels/ConfirmPaymentDetailsViewModel.swift
```diff
@@ -9,6 +9,7 @@ struct ConfirmPaymentDetailsViewModel {
     let session: WalletSession
     let config: Config
     let server: RPCServer
+    let keystore: Keystore
```
vikmeup: we have to pass `WalletInfo` in here, instead of keystore

Trust/Transfer/ViewModels/ConfirmPaymentDetailsViewModel.swift
```diff
@@ -53,6 +56,26 @@ struct ConfirmPaymentDetailsViewModel {
...
+
+    var currentWalletAddress: String {
+        return currentWallet?.address.description ?? ""
+    }
+
+    var currentWalletDescriptionString: String {
+        return currentWalletName + " " + ("(\(currentWalletAddress))")
+    }
+
     var paymentFromTitle: String {
```
vikmeup: You should use `paymentFromTitle` property

Trust/Transfer/ViewModels/ConfirmPaymentDetailsViewModel.swift
```diff
@@ -53,6 +56,26 @@ struct ConfirmPaymentDetailsViewModel {
         return transaction.gasLimit
     }
 
+    var walletNameTitle: String {
+        return NSLocalizedString("confirmPayment.walletTitle.label.title", value: "From Wallet", comment: "")
+    }
+
+    private var currentWallet: WalletInfo? {
+        return keystore.recentlyUsedWallet ?? keystore.wallets.first
```
vikmeup: pass current wallet and then use it here.

Trust/Transfer/ViewModels/ConfirmPaymentDetailsViewModel.swift
BloodyPixy: WalletSession seems to contain the needed wallet info. Didn't notice it earlier.

Trust/Localization/en.lproj/Localizable.strings
```diff
@@ -212,7 +212,7 @@
 "configureTransaction.dataField.label.title" = "Data (Optional). %@";
 "Nonce" = "Nonce";
 "transaction.recipient.label.title" = "Recipient";
-"transaction.sender.label.title" = "Sender";
+"transaction.sender.label.title" = "From";
```
vikmeup: do not change this.

Trust/Transfer/ViewModels/ConfirmPaymentDetailsViewModel.swift
```diff
@@ -56,28 +53,20 @@ struct ConfirmPaymentDetailsViewModel {
...
+        return session.account.address.description
     }
 
     var currentWalletDescriptionString: String {
         return currentWalletName + " " + ("(\(currentWalletAddress))")
     }
 
     var paymentFromTitle: String {
-        return NSLocalizedString("transaction.sender.label.title", value: "Sender", comment: "")
+        return NSLocalizedString("transaction.sender.label.title", value: "From", comment: "")
```
vikmeup: same here. Use `R.string.localized`

Trust/Localization/de.lproj/Localizable.strings
```diff
@@ -216,6 +216,7 @@
 "Are you sure you would like to delete?" = "Sind Sie sicher, dass Sie löschen möchten?";
 "confirmPayment.amount.label.title" = "Betrag";
 "confirmPayment.dapp.label.title" = "DApp";
+"confirmPayment.walletTitle.label.title" = "Aus Brieftasche";
```
vikmeup: can we remove this? do not pull strings unless

Trust/Localization/en.lproj/Localizable.strings
```diff
@@ -212,10 +212,11 @@
...
 "Are you sure you would like to delete?" = "Are you sure you would like to delete?";
 "confirmPayment.amount.label.title" = "Amount";
 "confirmPayment.dapp.label.title" = "DApp";
+"confirmPayment.walletTitle.label.title" = "From Wallet";
```
vikmeup: same here

TrustTests/Transfer/ViewModels/ConfirmPaymentDetailsViewModelTests.swift
```diff
@@ -0,0 +1,39 @@
...
+import RealmSwift
+import BigInt
+@testable import Trust
+
+class ConfirmPaymentDetailsViewModelTests: XCTestCase {
+
+    func testActionButtonTitleOnSignAndSend() {
+        let bigInt = BigInt("11274902618710000000000")!
+        
+        let transaction = PreviewTransaction(
```
vikmeup: make factory for `PreviewTransaction`, so you don't need to specify so much

vikmeup:
@BloodyPixy this won't work, you need to pass current `Account`, because you won't be able to see correct wallet if you use a main wallet (multi-coin.)

Issue #830, opened 2018-07-16
Require passcode on sending transactions


vikmeup:
Make it optional in the settings.

@madcake let's do the same for Android. I would put this into Security settings.

PR #867, opened 2018-07-30
Add generic passwork checker before sending transaction

BloodyPixy:
Referencing #830, I will create new coordinator for this one, and reuse the `LockEnterPasscodeViewController`.
I would like to get some info before finishing:

1. Should we include biometric confirm, as well as passcode?
2. Should we include 'Cancel' button for passcode screen? If so, should we store the current number of unsuccessful attempts?
3. If the number of unsuccessful attempts is exceeded, should we log the user out?

### DIFF
--- a/Trust/Transfer/ViewControllers/ConfirmPaymentViewController.swift
+++ b/Trust/Transfer/ViewControllers/ConfirmPaymentViewController.swift
@@ -23,6 +23,9 @@ class ConfirmPaymentViewController: UIViewController {
     lazy var sendTransactionCoordinator = {
         return SendTransactionCoordinator(session: self.session, keystore: keystore, confirmType: confirmType, server: server)
     }()
+    lazy var confirmEnterPasscodeCoordinator: LockEnterPasscodeCoordinator = {
+        return LockEnterPasscodeCoordinator(model: LockEnterPasscodeViewModel())
+    }()
     lazy var submitButton: UIButton = {
         let button = Button(size: .large, style: .solid)
         button.translatesAutoresizingMaskIntoConstraints = false
@@ -207,13 +210,19 @@ class ConfirmPaymentViewController: UIViewController {
     }
 
     @objc func send() {
-        self.displayLoading()
+        confirmEnterPasscodeCoordinator.start { [weak self] (success, _) in
+            if success {
+                self?.displayLoading()
 
-        let transaction = configurator.signTransaction
-        self.sendTransactionCoordinator.send(transaction: transaction) { [weak self] result in
-            guard let `self` = self else { return }
-            self.didCompleted?(result)
-            self.hideLoading()
+                guard let transaction = self?.configurator.signTransaction else {
+                    return
+                }
+                self?.sendTransactionCoordinator.send(transaction: transaction) { [weak self] result in
+                    guard let `self` = self else { return }
+                    self.didCompleted?(result)
+                    self.hideLoading()
+                }
+            }
         }
     }
 }
--- a/Trust/Lock/Coordinators/LockEnterPasscodeCoordinator.swift
+++ b/Trust/Lock/Coordinators/LockEnterPasscodeCoordinator.swift
@@ -2,6 +2,8 @@
 
 import UIKit
 
+typealias UnlockResult = ((_ success: Bool, _ bioUnlock: Bool) -> Void)
+
 final class LockEnterPasscodeCoordinator: Coordinator {
     var coordinators: [Coordinator] = []
     let window: UIWindow = UIWindow()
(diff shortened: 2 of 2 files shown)


vikmeup:
@BloodyPixy I have added a new coordinator, forgot that name, take a look in the Lock coordinators. I haven’t done much tho.

BloodyPixy:
@vikmeup 'Make sure use correct view model with messanging' - didn't get this one

vikmeup:
@BloodyPixy You need to pass view model that will describe actions!

Trust/Lock/ViewModels/LockEnterPasscodeViewModel.swift
```diff
@@ -2,8 +2,14 @@
 
 import UIKit
 
-final class LockEnterPasscodeViewModel: LockViewModel {
-    let initialLabelText =  NSLocalizedString("lock.enter.passcode.view.model.initial", value: "Enter your passcode.", comment: "")
-    let tryAfterOneMinute =  NSLocalizedString("lock.enter.passcode.view.model.try.after.one.minute", value: "Try after 1 minute.", comment: "")
-    let loginReason = NSLocalizedString("lock.enter.passcode.view.model.touch.id", value: "Logging in with Touch ID", comment: "")
+final class LockEnterPasscodeViewModel: LockViewModel, EnterPasscodeViewModelInterface {
```
vikmeup: why do we need to duplicate this text for each view model?

vikmeup: Also use `R.string.localized`

Trust/Transfer/ViewControllers/ConfirmPaymentViewController.swift
```diff
@@ -207,12 +212,24 @@ class ConfirmPaymentViewController: UIViewController {
...
     @objc func send() {
-        self.displayLoading()
+        guard let needsPasscodeCheck = PreferencesController()
+            .get(for: PreferenceOption.isPasscodeTransactionLockEnabled.key) as? Bool else {
+            return
+        }
 
-        let transaction = configurator.signTransaction
+        if needsPasscodeCheck {
+            delegate?.confirmPaymentControllerNeedsAuthentication(self)
```
vikmeup: is there a way to check this on coordinator level?

Trust/Lock/ViewModels/LockEnterPasscodeViewModel.swift
BloodyPixy: Check updated code

Trust/Transfer/Coordinators/ConfirmCoordinator.swift
```diff
@@ -76,3 +81,22 @@ final class ConfirmCoordinator: RootCoordinator {
         delegate?.didCancel(in: self)
     }
 }
+
+extension ConfirmCoordinator: ConfirmPaymentAuthenticationDelegate {
+    func confirmPaymentControllerNeedsAuthentication(_ controller: ConfirmPaymentViewController) {
+        guard let needsPasscodeCheck = PreferencesController()
```
vikmeup: This means you can modify user defaults to enable this feature? 

It needs to be stored on keychain level then

Trust/Settings/ViewControllers/SettingsViewController.swift
```diff
@@ -58,6 +59,23 @@ final class SettingsViewController: FormViewController, Coordinator {
         }
     }()
 
+    lazy var passcodeTransactionLockRow: SwitchRow = {
+        return SwitchRow(Values.passcodeTransactionLockRow) { [weak self] in
+            $0.title = self?.viewModel.verifyTransactionsWithPasscodeTitle
+            $0.value = self?.viewModel.isPasscodeTransactionLockEnabled
+            $0.hidden = Condition.function([Values.passcodeRow], { form in
+                return !((form.rowBy(tag: Values.passcodeRow) as? SwitchRow)?.value ?? false)
```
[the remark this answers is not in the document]
BloodyPixy: you mean for passcodeRow, and use it instead of this  
`!((form.rowBy(tag: Values.passcodeRow) as? SwitchRow)?.value ?? false)`
like this 
`!(self?.passcodeRow?.value ?? false)`,
did I get you right?

Trust/Settings/ViewControllers/SettingsViewController.swift
```diff
@@ -328,6 +349,12 @@ final class SettingsViewController: FormViewController, Coordinator {
         self.autoLockRow.reload()
     }
 
+    private func updatePasscodeTransactionLockRow(enabled: Bool) {
+        self.viewModel.isPasscodeTransactionLockEnabled = enabled
```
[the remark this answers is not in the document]
BloodyPixy: Done

Trust/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -48,6 +48,25 @@ struct SettingsViewModel {
...
+            }
+            return boolValue
+        }
+        set {
+            PreferencesController().set(value: newValue, for: PreferenceOption.isPasscodeTransactionLockEnabled)
+        }
+    }
+
+    var verifyTransactionsWithPasscodeTitle: String {
+        return NSLocalizedString("settings.lockTransactions.label.title", value: "Authenticate Transactions", comment: "")
```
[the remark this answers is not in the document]
BloodyPixy: Done

BloodyPixy:
I decided to put it inside of the lock. Seems logical to me, as we use Lock for keychain work with a passcode, and transaction authorization is directly connected to it. Good thing would be to have a new icon for this option in settings, as for now it uses same icon as Passcode app lock. 

It works like this: 
1. User enables app passcode protection and sets the passcode 
2. New line in table appears with a switch to authorize transactions
3. User can turn it 'on' 
4. If user turns App passcode lock off, transaction authorization will turn itself off as well. 
5. When user  performs a transaction with Authorization turned on, app checks for this value, and presents passcode protection view. 
7. If the passcode is correct, transaction is executed.

vikmeup:
Let's hold off on this one. It still requires UX/UI work to figure this out.

Issue #864, opened 2018-07-30
Add link to support center


vikmeup:
From the settings. 
Design 
![artboard 3](https://user-images.githubusercontent.com/725247/43599903-b2323560-9691-11e8-83e9-f23044e865ca.png)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #876, opened 2018-08-02
Add link to help center

ArtemGoryunov:
Issue link #864


vikmeup:
@ArtemGoryunov it might be great to delete unused icons from the assets.

do you know any good script to find unused variables, functions, files and etc?

PR #878, opened 2018-08-03
Bug/fix for lock bruteforce

BloodyPixy:
Please, take a look at the code.


TrustTests/Lock/View Controllers/LockEnterPasscodeViewControllerTests.swift
```diff
@@ -0,0 +1,11 @@
+// Copyright DApps Platform Inc. All rights reserved.
+
+import XCTest
+
+class LockEnterPasscodeViewControllerTests: XCTestCase {
```
vikmeup: can you add test case? Also there is a space in the file name, can you remove its?

OlegGordiichuk:
@BloodyPixy hmm nice work. Do you select same approach that we have discussed during Skype call ? (About passcode and touch flow)

BloodyPixy:
@OlegGordiichuk yes, as I proposed, that's how it will work.

Trust/Lock/ViewControllers/LockPasscodeViewController.swift
```diff
@@ -6,11 +6,12 @@ class LockPasscodeViewController: UIViewController {
     var willFinishWithResult: ((_ success: Bool) -> Void)?
     let model: LockViewModel
     var lockView: LockView!
-    let lock = Lock()
+    let lock: Lock!
```
vikmeup: can we avoid force unwrap?

Issue #875, opened 2018-08-02
Link "Settings → Telegram Group" should open channel according to user's locale


ArtemGoryunov:
List of available Telegram Groups are here https://github.com/TrustWallet/docs/blob/master/telegram_communities.md

PR #879, opened 2018-08-03
Link "Settings → Telegram Group" should open channel according to user's locale

ArtemGoryunov:
Referencing to #875


Trust/Settings/Types/ServiceProvider.swift
```diff
@@ -36,7 +36,9 @@ enum URLServiceProvider {
         case .twitter:
             return URL(string: "twitter://user?screen_name=\(Constants.twitterUsername)")!
         case .telegram:
-            return URL(string: "tg://resolve?domain=\(Constants.telegramUsername)")
+            let languageCode = NSLocale.preferredLanguageCode ?? ""
```
vikmeup: can you move this code into a function? it's been reused below

Issue #885, opened 2018-08-05
Crash: Signing on watch address via browser


vikmeup:
<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

darlhz:
@vikmeup I think I've fixed the issue. please have a look :)

PR #887, opened 2018-08-05
fix crash when signing on watch address via browser #885

darlhz:


vikmeup:
@darlhz looks good! Even tho we need to show an alert before it gets here and say that this is watch only wallet, I will make that change.

PR #920, opened 2018-09-12
Remove extra method case

omatty198:
I found extra method case.


vikmeup:
Why extra? It’s needed

omatty198:
@vikmeup 
I'm sorry, Is it for future use?

